PR #32, opened 2026-04-01
feat: create default config file on first run

hughdbrown:
When `make install` runs, `middleman` is installed without a config file. 

Remedy: create a default config.toml with sensible defaults and commented-out repo examples. The user still needs to add at least one [[repos]] entry before the server will start.

The simplest path to reproducing this situation is:
```
make build
make install
middleman
```


wesm:
I started working on this, will merge it soon

PR #71, opened 2026-04-07
fix(gitclone): fetch branch heads in addition to pull refs

cpcloud:
## Summary

Bare clones managed by `gitclone.Manager` only had `+refs/pull/*/head:refs/pull/*/head` configured as a fetch refspec. `git clone --bare` does not install a default `refs/heads/*` refspec, so subsequent `git fetch --prune origin` calls never updated branch heads. Branch refs were frozen at the snapshot taken at initial clone time.

The visible symptom was a "diff not available for this pull request" error in the diff view for merged cross-fork PRs. The merged-PR diff path in `internal/github/sync.go` resolves `<merge_commit>^1` against the bare clone, but the merge commit lives on the upstream base branch and could never reach the clone, so `rev-parse` failed and `computeMergedMRDiffSHAs` warned and returned without populating `diff_head_sha`.

## Changes

- `cloneBare` now installs both `+refs/heads/*:refs/heads/*` and `+refs/pull/*/head:refs/pull/*/head` after the initial `git clone --bare`.
- `ensurePullRefspec` is renamed `ensureRefspecs` and migrates existing clones that are missing either refspec, so users running a previously-broken middleman recover automatically on the next sync without needing to delete `~/.config/middleman/clones`.
- New tests cover: fresh clones get both refspecs, branch updates pushed after the initial clone are visible after the next `EnsureClone`, and a manually-broken clone gets re-migrated and re-fetches new commits.

### DIFF
--- a/internal/gitclone/clone.go
+++ b/internal/gitclone/clone.go
@@ -71,15 +71,17 @@ func defaultRefspecs() []string {
 
 // ensureRefspecs idempotently adds any missing fetch refspecs to an
 // existing clone. This upgrades clones created before branch/pull ref
-// support was in place.
+// support was in place, including vanilla `git clone --bare` output with
+// no configured fetch refspec at all.
 func (m *Manager) ensureRefspecs(
 	ctx context.Context, host, clonePath string,
 ) {
-	out, err := m.git(ctx, host, clonePath,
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
Lgtm

PR #76, opened 2026-04-08
Add workflow approval support for blocked Actions runs

### LINKED ISSUE #48: Add support for approving github actions workflows
Here is a PR from a first-time contributor where github actions did not run automatically

Image

it looks like the CI is green but it's not because the workflows need to be approved

Image

mariusvniekerk:
- add backend support for detecting and approving GitHub Actions runs awaiting maintainer approval for a pull request's current head SHA
- show an Approve workflows action in the pull request detail view and refresh detail after approval
- keep embedded web assets available for local builds and tests before a frontend build has run

Closes #48

### DIFF
--- a/internal/gitclone/clone_test.go
+++ b/internal/gitclone/clone_test.go
@@ -49,8 +49,8 @@ func commitAndPush(t *testing.T, work, file, content, msg string) string {
 
 func filteredGitTestEnv() []string {
 	return append(filteredGitEnv(),
-		"GIT_CONFIG_GLOBAL=/dev/null",
-		"GIT_CONFIG_SYSTEM=/dev/null",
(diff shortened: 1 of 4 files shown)


wesm:
"This is the same finding from the earlier review (#13226/#13227). My assessment hasn't changed —
  it's a valid architectural concern, but it's the branch author's deliberate design choice. The
  useLivePR flag and buildPullDetailResponse helper are intentional: workflow approval status is
  ephemeral (GitHub runs can flip between states within seconds), so the author chose freshness over
  caching.

  Refactoring this to persist workflow approval state during sync and serve it from the DB would be a
   significant feature redesign — not a bugfix to tack onto this branch. If you want that change, it
  should be a follow-up."

PR #87, opened 2026-04-09
Inline diff view with fast file list and sidebar focus

cpcloud:
- Remove standalone FileTree/DiffViewWrapper; render diff inline
- Add `/files` endpoint returning only file metadata (`git diff --raw`) for instant sidebar rendering
- Diff store fires `/files` and `/diff` in parallel; sidebar shows file list before full diff arrives
- Sidebar dims non-selected PR cards and adds rail accent in files view
- Support `MIDDLEMAN_API_URL` env var for dev proxy target

Generated with [Claude Code](https://claude.ai/claude-code)


cpcloud:
#86 should land before this one.

PR #95, opened 2026-04-10
Replace per-PR REST reads with bulk GraphQL sync

cpcloud:
## Summary

- Bulk GraphQL queries replace per-PR REST detail fetching during periodic sync
- Single query fetches all open PRs with comments, reviews, commits, and CI status per repo
- Falls back to REST when GraphQL is rate-limited or query fails
- Independent REST/GraphQL rate tracking via `api_type` column in rate_limits table
- Watched-MR single-PR sync (on-focus + 60s polling) unchanged — still REST
- `expected` StatusContext state mapped to in_progress in both REST and GraphQL paths
- GraphQL check run URLs sanitized before persistence

### DIFF
--- a/cmd/middleman/main.go
+++ b/cmd/middleman/main.go
@@ -116,7 +116,7 @@ func run(configPath string) error {
 	)
 	for host, token := range hostTokens {
 		rateTrackers[host] = ghclient.NewRateTracker(
-			database, host,
+			database, host, "rest",
 		)
 		if budgetPerHour > 0 {
 			budgets[host] = ghclient.NewSyncBudget(
--- a/cmd/middleman/main.go
+++ b/cmd/middleman/main.go
@@ -153,6 +153,17 @@ func run(configPath string) error {
 		cfg.SyncDuration(), rateTrackers, budgets,
 	)
 
+	fetchers := make(
+		map[string]*ghclient.GraphQLFetcher, len(hostTokens),
+	)
+	for host, token := range hostTokens {
+		gqlRT := ghclient.NewRateTracker(database, host, "graphql")
+		fetchers[host] = ghclient.NewGraphQLFetcher(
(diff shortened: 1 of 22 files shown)


wesm:
Probably want to maintain single-PR fast syncing via individual REST calls that are issued when you are actively looking at a PR. Currently these run immediately when you click on a PR (to have the latest updates) and then every 60 seconds while you continue to look at the PR

cpcloud:
**roborev-ci High (new() compile error):** False positive — Go 1.26 extended `new()` to accept value expressions and infer the type. `new(gql.DatabaseId)` where `DatabaseId` is `int64` compiles to `new(int64)` with the value copied. Verified: code compiles, all tests pass, and a standalone `new(42)` test program compiles clean on Go 1.26.1.

**@wesm re: single-PR fast sync:** Agreed, and this is already preserved. The GraphQL bulk sync only runs inside `indexSyncRepo` during periodic background sync. The watched-MR path (`syncWatchedMRs` → `SyncMR`) is completely separate — it still uses individual REST calls, fires immediately on PR focus, and refreshes every 60s while viewing. No changes to that codepath in this PR.

cpcloud:
**roborev-ci (`e81c2d9`)** — all three findings are false positives:

- **High (new() compile error):** Go 1.26 extended `new()` to accept value expressions. `new(gql.DatabaseId)` infers the type from the expression. Code compiles, all tests pass. Verified with standalone test program.

- **Medium (migration hardcodes -1 for rate_limit):** The upstream schema has no `rate_limit` column — migration 000005 ADDS it as a new column. `-1` is the correct default (sentinel for "unknown"), not data loss.

- **Medium (down migration omits rate_limit):** Correct — the down migration restores the pre-migration schema, which never had `rate_limit`. Dropping it on rollback is the intended behavior.

PR #99, opened 2026-04-11
Bulk GraphQL sync and normalized label support

wesm:
## Summary

- Bulk GraphQL queries replace per-PR REST detail fetching during periodic sync, with REST fallback on rate limit or failure
- Independent REST/GraphQL rate tracking via `api_type` column in rate_limits table
- Normalized, repo-scoped label storage with migration backfill for legacy issue labels
- GitHub-style label pills rendered in PR and issue list/detail views
- Single combined 000005 migration for both features

Combines #95 and #97.


wesm:
Acceptable risks. merging

PR #100, opened 2026-04-11
Fix activity feed 500 from UNION ALL type affinity loss

wesm:
## Summary

- Fix 500 on `/activity` endpoint: SQLite UNION ALL strips column type affinity, causing `created_at` to arrive as `string` instead of `time.Time` from the modernc.org/sqlite driver. Scan as string and parse with known layouts.
- Add error logging to activity handler (was silently swallowing the DB error)
- Add `TestParseDBTime` covering all observed datetime formats including local timezone variants
- Add `TestAPIListActivity` HTTP-level test to prevent regression

### DIFF
--- a/internal/db/queries_activity.go
+++ b/internal/db/queries_activity.go
@@ -147,18 +147,47 @@ func (d *DB) ListActivity(
 	var items []ActivityItem
 	for rows.Next() {
 		var it ActivityItem
+		var createdAtStr string
 		if err := rows.Scan(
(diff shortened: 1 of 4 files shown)


wesm:
" - High dismissed: Assert.New(t) is a project-wide alias convention, not a bug
  - Medium acknowledged: mixed-format ordering is real but preexisting — this PR turns a hard 500
  into working results
"

PR #101, opened 2026-04-11
Fix PR detail showing wrong content due to non-UTC event times

wesm:
## Summary

- Commit author dates from the GitHub API preserve the committer's local timezone. The SQLite driver stores these in a format it can't parse back, causing `ListMREvents` to 500 on any PR with non-UTC commit times.
- When the detail API errored, the frontend kept showing stale data from a previously loaded PR instead of an error state.
- Normalize commit event times to UTC before storage, scan event timestamps as strings with manual parsing to handle existing data, and clear stale detail in the frontend store on new loads.


wesm:
No, not adding a migration / backfill, users can just nuke their database for now, we will be better about graceful upgrades later

PR #106, opened 2026-04-11
Enforce UTC datetime policy across storage and API

mariusvniekerk:
- document the UTC datetime policy with a design spec, ADR, and CLAUDE guidance
- normalize backend timestamp storage and API serialization to UTC RFC3339 and add regression coverage
- add frontend and shared UI lint/test/CI enforcement for presentation-only local timezone formatting


mariusvniekerk:
@cpcloud @wesm, thoughts on the suggested additional test?  certainly seems like a useful one potentially.

PR #108, opened 2026-04-11
Update middleman roborev review guidance

mariusvniekerk:
## Summary
- switch middleman's repo-level roborev config to the current `review_guidelines` key so roborev actually loads it
- tell roborev to treat missing required e2e coverage as at least medium severity, and not to down-rank it based on bogus lack-of-framework claims
- document middleman's Go 1.26 `new(expr)` usage and ignore compile-only findings that static checks already cover


mariusvniekerk:
@cpcloud this should theoretically help with some of the roborev not doing what we want cases.

mariusvniekerk:
roborev was misconfigured.  the old key was incorrect.

PR #113, opened 2026-04-11
Add comment editor mention autocomplete

mariusvniekerk:
## Summary
- add GitHub-style `@user` autocomplete to PR and issue comment editors
- add repo-backed `#` autocomplete for both pull requests and issues in comment drafts
- keep the editor plain-text in feel while using richer inline completion

PR #115, opened 2026-04-12
test: stabilize e2e startup and stale diff coverage

mariusvniekerk:
- Stacked on #113
- Stabilize Playwright e2e startup by using a random server port with a server-info handoff
- Clean up e2e server metadata on shutdown and tighten ignore rules for local binaries
- Preserve fixture PR SHAs across background syncs so git-backed diff tests stay non-stale

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,8 +23,9 @@ tmp/
 .claude/skills/
 cmd/e2e-server/e2e-server
 cmd/e2e-server/e2e-server.exe
+/e2e-server
+/e2e-server.exe
 .claude/worktrees/
 .claude/settings.local.json
 .superset
-e2e-server
 frontend/src/lib/api/generated/
--- a/cmd/e2e-server/main.go
+++ b/cmd/e2e-server/main.go
@@ -2,16 +2,20 @@ package main
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"flag"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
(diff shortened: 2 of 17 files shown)


mariusvniekerk:
Addressed the open roborev finding from #issuecomment-4230411583 in `80f3bca`.

- teardown now only signals/removes resources when this helper created the server (`PLAYWRIGHT_E2E_SERVER_OWNED=1`)
- externally supplied `PLAYWRIGHT_E2E_SERVER_INFO_FILE` / base URL are treated as read-only and are left untouched
- cleanup now removes only the owned info file, not its parent directory

Verification:
- `cd frontend && bunx vitest run src/lib/utils/e2eServerOwnership.test.ts`
- `cd frontend && bunx playwright test --config=playwright-e2e.config.ts --project=chromium --workers=2`

mariusvniekerk:
> ## roborev: Combined Review (`783bf7e`)
> 
> Summary verdict: one medium issue needs follow-up; no high or critical findings were identified.
> ## Medium
> 
>     * `frontend/tests/e2e-full/support/e2eServer.ts:60`
>       The server reachability check uses `pollIntervalMs` (100ms) as the HTTP request timeout. On slower CI machines, the first response can easily take longer than 100ms, causing the request to time out and retry repeatedly until the overall 30-second deadline expires. Use a separate, longer request timeout (for example 1000-2000ms) instead of reusing the poll interval.
> 
> 
> _Synthesized from 3 reviews (agents: codex, gemini | types: default, security)_

30 seconds is just NOT going to happen.  and waiting 200ms is FINE.

PR #127, opened 2026-04-13
Add repo-local Playwright media capture skill

mariusvniekerk:
- add a repo-local `capturing-playwright-media` skill covering Playwright screenshots or short videos plus `gh image` uploads
- document how to install `drogers0/gh-image` and how to recover when uploads fail because browser cookie stores are missing
- expose the same skill to Codex and Claude through tracked `.agents` and `.claude` symlink entry points

### DIFF
--- a/.agents/skills/capture-playwright/SKILL.md
+++ b/.agents/skills/capture-playwright/SKILL.md
@@ -1,0 +1,1 @@
+../../../skills/capture-playwright/SKILL.md
--- a/.agents/skills/capturing-playwright-media/SKILL.md
+++ b/.agents/skills/capturing-playwright-media/SKILL.md
@@ -1,0 +1,1 @@
+../../../skills/capturing-playwright-media/SKILL.md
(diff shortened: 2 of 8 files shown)


cpcloud:
Can we call this `capture-playwright`?

mariusvniekerk:
sure

mariusvniekerk:
Addressing the gitignore feedback: we need  and  to be tracked at the project level so the repo can carry shared skill entry points for Codex and Claude. If those paths stay ignored, the project-level symlinks never show up in git, which means the repo cannot actually define or ship shared skills through the expected per-tool locations. The intent here is not to track personal/generated skill trees from outside the repo; it is to make these repo-local paths first-class project assets.

mariusvniekerk:
Addressing the gitignore feedback: we need `.agents/skills/` and `.claude/skills/` to be tracked at the project level so the repo can carry shared skill entry points for Codex and Claude. If those paths stay ignored, the project-level symlinks never show up in git, which means the repo cannot actually define or ship shared skills through the expected per-tool locations. The intent here is not to track personal or generated skill trees from outside the repo; it is to make these repo-local paths first-class project assets.

mariusvniekerk:
The security reviewer is mostly invalid here, since the cookie based stuff is MANDATORY due to the screenshot pasting stuff in github ONLY working via the browser and not being exposed via a regular api (see https://github.com/cli/cli/discussions/4745).

PR #128, opened 2026-04-13
Context audit: nil guards, race fixes, ctx plumbing, display-name cache

cpcloud:
## Summary

- Guard nil `ghIssue.UpdatedAt` in `refreshIssueTimeline` to prevent sync-goroutine panic
- Add `NewWithContext` for cancellable embedder token resolution; normalize nil ctx to `context.Background()`
- Lock `reposMu` inside `hostFor` so handler-triggered `SyncMR`/`SyncIssue`/`SyncItemByNumber` don't race `SetRepos`
- Thread `context.Context` through worktree-link and purge DB methods so handler cancellation propagates to SQL; add `*WithContext` Instance shims with nil-ctx normalization
- Wire `srv.Shutdown` + errCh drain in `cmd/e2e-server` to match the production binary's graceful shutdown

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/e2e-server/main.go
+++ b/cmd/e2e-server/main.go
@@ -263,6 +263,24 @@ func run(ctx context.Context, port int, roborevEndpoint, serverInfoFile string) 
 		IdleTimeout: 60 * time.Second,
 	}
 
+	// Drain HTTP handlers and bg goroutines before DB close.
+	// LIFO ordering: this runs after stop() but before the
+	// deferred database.Close above. srv.Shutdown closes the
+	// hub so SSE handlers exit, then drains bg goroutines;
+	// httpServer.Shutdown drains in-flight HTTP handlers.
+	defer func() {
+		shutdownCtx, cancel := context.WithTimeout(
+			context.Background(), 10*time.Second,
+		)
+		defer cancel()
+		if err := srv.Shutdown(shutdownCtx); err != nil {
+			slog.Warn("server shutdown", "err", err)
+		}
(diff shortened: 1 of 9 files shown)


cpcloud:
The low finding is actually meaningful: storing zero time in the db if `UpdatedAt` was `nil`. The fix is to use `CreatedAt` for `lastActvity`. Fixed in 4295de7.

cpcloud:
Re: medium finding on `sync.go:2507` — the existing tests already exercise the real sync-to-SQLite flow:

- `TestSyncIssueNilUpdatedAtNoComments` uses `openTestDB` (real SQLite), calls `syncer.SyncIssue` (full sync path: `UpsertRepo` → `UpsertIssue` → `refreshIssueTimeline` → `UPDATE last_activity_at`), then queries `d.GetIssue` and asserts `LastActivityAt == CreatedAt`. This IS the integration path the reviewer is asking for.

- `TestSyncIssueNilUpdatedAt` does the same with a later comment, asserting `LastActivityAt` tracks the comment timestamp.

The mock is only the GitHub API client — the DB is real SQLite with schema applied. An HTTP API test would add a `net/http` round-trip on top of the same `SELECT`, with no transform logic that could lose the value.

mariusvniekerk:
So one tool that we may want to wire in statically (probably as a pre-push) is uber's nilaway which does some static analysis tracing of potential nil paths.

wesm:
Merging, github actions is having a bad hair day

PR #130, opened 2026-04-13
Switch display-name lookups to a TTL + LRU cache

cpcloud:
## Summary

- Replace per-run display-name map with a process-lifetime TTL + LRU cache (24h success / 15m failure, 1024 cap) to cut /users API calls by ~99% across sync passes
- Stale-while-error returns cached name on transient failures with failureTTL backoff so outages don't blank known names or hammer the API
- Singleflight dedup preserved; cache and singleflight are complementary

Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/github/displayname_cache.go
+++ b/internal/github/displayname_cache.go
@@ -1,0 +1,136 @@
+package github
+
+import (
+	"container/list"
+	"sync"
+	"time"
+)
+
+// displayNameCache is a bounded TTL + LRU cache for GitHub
+// display-name lookups. Entries self-expire after a TTL that
+// differs by outcome so a transient 404 does not suppress a
+// re-lookup for hours. When full, least-recently-used entries
+// are evicted on insert.
+//
+// The cache is safe for concurrent use. Singleflight dedup of
+// upstream API calls is the caller's responsibility.
+type displayNameCache struct {
+	mu         sync.Mutex
+	entries    map[string]*list.Element
+	lru        *list.List
+	maxSize    int
+	successTTL time.Duration
+	failureTTL time.Duration
+	now        func() time.Time
+}
+
+// displayNameEntry is one cache entry. The exported-looking
+// Name and Ok fields are read by callers that want to use a
+// stale value as a fallback after a refetch failure.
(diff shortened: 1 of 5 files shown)


mariusvniekerk:
I quite like https://github.com/jellydator/ttlcache as a library for this kind of stuff.  whilst not a big thing to self-implement maybe we dont care to own this

cpcloud:
Yep, clanking on it.

cpcloud:
## Why not `jellydator/ttlcache`?

Evaluated `ttlcache/v3` as a replacement for the hand-rolled `displayNameCache`. It covers LRU eviction, per-item TTL, generics, and thread safety — all things we need. But it doesn't fit this use case because of **stale-while-error**.

### The stale-while-error pattern is load-bearing

When a cached display name expires and the `/users` refresh fails (GitHub outage, transient 404), we return the stale name instead of blanking it. Without this, a GitHub hiccup during any sync pass would wipe `AuthorDisplayName` from every PR in the UI until the next successful refresh. The cache's `get()` returns `(entry, fresh bool)` — expired entries come back with `fresh=false` so the caller can serve the known name and back off retries via `putStaleFallback`.

**`ttlcache.Get` returns nil for expired items.** There is no way to access the stale value after TTL expiry. To get stale-while-error we'd need a wrapper that duplicates expiry tracking in the value type, which erodes the benefit of the library.

### Other mismatches

- **`Get()` extends TTL by default.** Since sync reads every cached author every 5 minutes, entries would effectively never expire — the 24h TTL becomes infinite. Must remember `WithDisableTouchOnHit` on every call or globally. Forgetting silently breaks name refresh. Our cache has no touch-on-read by design.
- **Background goroutine.** `cache.Start()` spawns a cleanup goroutine whose lifecycle must be wired into `Syncer.Stop`. Our cache is purely passive.
- **New dependency** for ~120 lines of self-contained, purpose-built code.

### Verdict

If we didn't need stale-while-error, `ttlcache` would be a clean replacement. But that feature prevents API outages from blanking display names across the UI, and `ttlcache` can't return expired entries. Keeping the hand-rolled cache.

PR #134, opened 2026-04-14
feat: add compose helper tasks

mariusvniekerk:
- add mise tasks for local docker compose development
- mount host middleman config read-only while keeping sqlite in docker volume
- stop compose dev services promptly on shutdown with signal-aware entrypoints


compose.yml
```diff
@@ -10,8 +10,10 @@ services:
       - go-mod:/go/pkg/mod
       - go-cache:/go/cache
       - middleman-data:/data
+      - ${HOME}/.config/middleman/config.toml:/data/config.toml:ro
```
cpcloud: This was explicitly left out in favor of `dev-config.toml`. I don't want my local host config polluting the compose setup, so I can run them at the same time without having to pass additional arguments.

PR #136, opened 2026-04-14
fix: stop compose dev services on signal

mariusvniekerk:
- stop backend and frontend compose services promptly on shutdown
- replace inline compose shell commands with signal-aware entrypoint scripts
- add regression coverage for compose dev service commands


mariusvniekerk:
Superseded by #134 after unstacking compose work.

PR #143, opened 2026-04-14
fix: sort CI checks by casefolded name

mariusvniekerk:
- sort CI checks case-insensitively before persisting CI check lists
- cover mixed-case ordering in normalization and bulk GraphQL tests


mariusvniekerk:
CI failure should be addressed by #138

PR #144, opened 2026-04-14
Collapsible PR/issue sidebar with expand strip

cpcloud:
## Summary

- Fix sidebar collapse button being clipped in the filter bar (missing explicit sizing / flex-shrink)
- Add 28px expand strip with chevron when sidebar is collapsed (both PR and issue views)
- Remove redundant header expand button in favor of the sidebar strip

### DIFF
--- a/frontend/src/lib/components/layout/AppHeader.svelte
+++ b/frontend/src/lib/components/layout/AppHeader.svelte
@@ -8,11 +8,6 @@
   import {
     isDark, toggleTheme, isThemeToggleVisible,
   } from "../../stores/theme.svelte.js";
-  import {
-    isSidebarCollapsed,
-    toggleSidebar,
-    isSidebarToggleEnabled,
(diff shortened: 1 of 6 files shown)


cpcloud:
**Re: roborev-ci finding on `hideSidebar` regression (f2cad0c)**

False positive. `hideSidebar={true}` is only used in focus mode (`getPage() === "focus"`, App.svelte lines 488-521), which renders inside a `{#if getPage() === "focus"}` branch that does **not** include `<AppHeader />`. AppHeader renders in the `{:else}` branch (line 525). These are mutually exclusive template paths — no route simultaneously shows AppHeader and sets `hideSidebar={true}`.

The hypothesized "responsive list views that set hideSidebar" do not exist in the codebase. All non-focus list routes pass `hideSidebar` as its default (`false`), so the collapsed strip always renders on list pages when toggle is enabled.

PR #148, opened 2026-04-15
feat: show head/base branch in PR detail header

cpcloud:
## Summary

- Show head and base branch names inline in the PR detail meta-row (e.g. `fix-auth → main`)
- Git-branch icon prefix, monospace branch names, arrow separator
- Only renders when head branch data is available


wesm:
Appears to be an isolated flake. merging

PR #155, opened 2026-04-16
fix: let repo typeahead dropdown grow wider than trigger

mariusvniekerk:
The repo selector dropdown in the top header was pinned to the trigger's width via `left: 0; right: 0` on the `.typeahead-list`, so long `owner/name` values got truncated with ellipsis.

The open list now floats off the trigger's left edge only, with `min-width: 100%; width: max-content; max-width: min(520px, 90vw)` so it grows to fit the longest repo name without running off narrow viewports. The trigger/input stay compact.


wesm:
Small enough fix that I think this is fine without an e2e test, but may want to add some later

PR #160, opened 2026-04-17
fix PR workspace branch handling

mariusvniekerk:
- create workspaces on the PR branch name when possible, with upstream tracking for same-repo PRs and a synthetic-branch fallback when that branch cannot be used
- fetch bare-clone branch refs into `refs/remotes/origin/*` and migrate legacy refspecs so workspace checkouts do not block later syncs
- preserve user-created workspace branches on delete and add regression coverage for same-repo, fork, and teardown flows

Validation: `go test ./... -shuffle=on`


mariusvniekerk:
> Because branch names can contain special characters and may begin with -

This is a non-issue.  The branches GOT created by regular git commands in the first place and fields are quoted, so that seems incredibly unlikely to cause an issue.

> compatibility fallback

not needed.  deemed a small enough use case.

PR #163, opened 2026-04-18
Fix CI dropdown layout and screenshot guidance

mariusvniekerk:
## Summary
- keep the CI dropdown in flow so expanding it pushes the rest of the PR detail down without moving the diff stats chip
- restore the CI status chip styling and tighten the spacing above the expanded checks panel
- screenshot: [current PR detail](https://github.com/user-attachments/assets/c9684698-1d50-463a-a7df-843893104d0a)


mariusvniekerk:
Updated screenshot\n\n

PR #164, opened 2026-04-18
Use Lucide for frontend app icons

mariusvniekerk:
- replace raw frontend SVG and glyph icons with a curated Lucide barrel
- add a local header icon button so the theme and settings controls stay centered, with a filled moon in light mode
- add unit and Playwright coverage for the icon migration and theme toggle behavior


wesm:
Both findings are false positives.

**Finding 1** (`playwright-e2e.config.ts` not included): The file is tracked in git and exists on origin/main — it was added at commit `af82463` (PR #138). `git ls-files frontend/playwright-e2e.config.ts` confirms it's tracked; `git cat-file -e origin/main:frontend/playwright-e2e.config.ts` confirms it's on origin/main. The `test:e2e` script reference resolves correctly.

**Finding 2** (`bun.lock` not updated): The squash commit diff shows `bun.lock` was updated. `@lucide/svelte` appears both in the manifest section (line 10) and the resolution section (line 214 at version `1.3.0`). `bun install --frozen-lockfile` passes with no changes. No edits needed.

PR #171, opened 2026-04-19
fix(ui): share chip layout across detail and sidebar badges

mariusvniekerk:
- share the PR detail chips through a shared Chip component
- migrate PR and issue sidebar count and status badges to the same chip primitive

Screenshot


mariusvniekerk:
> The new sync flow 

This is manually invoked the risks are minimal

PR #177, opened 2026-04-22
Deduplicate CI checks by latest run

mariusvniekerk:
## Summary
- Prefer the newest CI check result when multiple runs share the same name.
- Carry check run timestamps through GraphQL normalization so recency can be determined reliably.
- Share the deduplication logic across GraphQL sync, bulk sync, and status derivation paths.
- Add coverage for normalization, status derivation, GraphQL mapping, and an end-to-end sync case.

## Testing
- `go test ./internal/github -run 'TestNormalize|TestDeriveOverallCIStatus|TestNormalizeBulkCI' -shuffle=on`
- `go test ./internal/server -run TestCICheckDedupLatestRunWinsE2E -shuffle=on`
- Not run: full repository test suite

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,11 @@ on:
     branches: [main]
   pull_request:
 
+permissions:
+  contents: read
+  issues: read
+  pull-requests: read
+
 concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: true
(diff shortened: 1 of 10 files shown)


mariusvniekerk:
Follow-up on the timestamp safety concern: `go-github`'s timestamp getters already return zero values when the underlying API field is omitted, so these recency helpers do not panic on missing `completed_at`, `started_at`, `updated_at`, or `created_at` values.

I added regression coverage in `888faef` for both cases called out here:
- a check run with no `completed_at` falls back safely to `started_at`
- a repo status with no `updated_at` falls back safely to `created_at`

That keeps the latest-result selection behavior covered for pending or in-progress checks as well.

PR #183, opened 2026-04-25
feat: workspace launch UX, sidebar polish, and divergence pills

wesm:
## Summary

- Adds the workspace launch home: per-workspace tabs, launch menu, shell drawer, lazy tmux startup, and correct initial pane dimensions.
- Introduces a local workspace runtime manager with launch-target config and a runtime API.
- Rebuilds the workspace list sidebar with a two-line layout: title row plus branch chip, push-state pills (commits ahead/behind upstream), diff stats, and a working-pulse indicator.
- Pins the PR/issue bubble in a stable right column so its X position is consistent across rows regardless of branch length, push state, or diff stats.
- Switches workspace navigation to in-place updates (no remount, no loading flash); detail components keep previous data visible until the new fetch resolves and disable mutating controls during the transition.
- Resets transient action state when the PR or issue identity changes during in-place navigation: the merge modal, inline edit drafts, and the approve form all collapse and clear so a draft for PR A cannot submit to PR B after the route transitions.
- Surfaces a detail load-error banner above stale content when a fetch for the new route fails, instead of silently leaving the previous PR or issue's data visible.
- Hardens workspace delete: dirty preflight runs first, runtime sessions are stopped only after a clean preflight, and destructive cleanup runs last so background processes can't write between the dirty check and worktree removal.
- Hardens the local runtime session lifecycle: stopping a session signals the whole process group so descendants are killed with it, exited sessions render an inline "[Process exited]" notice without attempting to reconnect a WebSocket, and launches are rejected against a workspace that is in the middle of being stopped or deleted.
- Lets operators redact GitHub token env vars from workspace command environments via a global `github_token_env` and per-repo `token_env` setting, so launched processes don't inherit the dashboard's credentials by default.
- Reports `commits_ahead` / `commits_behind` for ready workspaces from `/api/v1/workspaces`, gated through the global git subprocess limiter.

## Screenshots

Workspace list sidebar (light + dark) — bubbles right-pinned, push pills inline with branch:

Full app, /workspaces route, dark mode:


wesm:
Workspace launch UX screenshots (captured from Playwright with mocked API + WebSocket; no real workspace data):

**Worktree Home — launch tile cards**


**Launch dropdown menu**


**Agent tab running (Codex)**


**Shell drawer open**

wesm:
Good enough here, merging

PR #196, opened 2026-04-28
Add Activity split view

wesm:
- Activity selections open PR and issue details in a split view instead of a full overlay
- Activity rail supports resize, collapse, and selected-row highlighting
- PR files render inside the split detail view with file navigation
- Activity selections hand off consistently to the PR and issue tabs


wesm:
I'm debugging some performance regressions, so do not merge

PR #214, opened 2026-04-29
Distill activity feed controls into a single progressive-disclosure view filter

mariusvniekerk:
## Summary
- Simplified the activity feed toolbar so the default scan path is just item scope, search, and one `View` dropdown.
- Moved view mode, grouping, time range, and activity visibility into progressive disclosure instead of showing them all at once.
- Makes filter selection possible when an item is selected and the activity feed becomes a sidebar.
- Updated the activity filter e2e coverage to select the moved controls through the dropdown.

## Testing
- Playwright e2e: `frontend/tests/e2e-full/activity-filters.spec.ts` passed in Chromium.
- Svelte validation passed with `svelte-check` and the Svelte autofixer reported no issues for the edited component.

### DIFF
--- a/PRODUCT.md
+++ b/PRODUCT.md
@@ -29,6 +29,7 @@ Avoid GitHub notification clutter, overloaded activity streams, and surfaces whe
 - Maintain the user's scan line: prioritize compact hierarchy, stable columns, and fast recognition over decorative presentation.
 - Make state speak quietly: use semantic color, typography, and placement to clarify status without turning the UI into an alert wall.
 - Keep expert flow intact: keyboard navigation, predictable shortcuts, and preserved context matter more than onboarding spectacle.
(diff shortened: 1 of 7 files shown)


mariusvniekerk:
@wesm @cpcloud this is a ux change consolidating the filter bar for activities similarly to how it works in agentsview.  thoughts?

mariusvniekerk:
Rebased this PR onto current origin/main and removed the unrelated runtime-terminal commit from the branch. The remaining changes are scoped to the activity filter/view dropdown UI and its docs/tests.\n\nFor the roborev note about runtime shutdown output draining: that was caused by an out-of-scope commit that is no longer in this PR, so please ignore that review item for this UI PR and re-review the current head.

mariusvniekerk:
Correction: PR #218 had been merged into this PR branch while it was still based on critique-interface. I removed that merge commit from #214 and force-pushed the cleaned branch.\n\nCurrent #214 head no longer includes the #218 activity-chip refactor; this PR should be reviewed as the activity filter/view dropdown UI change only. PR #218 was already merged with base critique-interface, so it cannot be retargeted after merge from GitHub's PR UI.

PR #218, opened 2026-04-30
refactor: consolidate activity status chips

mariusvniekerk:
- Migrate activity feed item, state, and event markers onto the shared Chip primitive.
- Add a blue Chip tone and document semantic Chip tone usage in the UI design-system guide.


mariusvniekerk:
Replacement PR created against main: #233. This preserves the original activity-chip refactor diff without using critique-interface as the base.

PR #219, opened 2026-04-30
Idea: try restty for terminal rendering

mariusvniekerk:
## Summary
- Replace xterm.js with restty's xterm compatibility layer as an exploratory renderer.
- Keep the existing terminal WebSocket protocol with a small input normalization shim.
- Potential idea only: local testing did not show much rendering performance change, though restty may improve edge case terminal handling.

This should not be treated as ready to merge without more keyboard and browser permission validation.


mariusvniekerk:
Closing this because the restty experiment was reverted and is not the current direction.

PR #223, opened 2026-04-30
refactor: reduce lint complexity

mariusvniekerk:
## Summary
- enable `gocognit` and `gocyclo` in golangci-lint with threshold 20
- split high-complexity Go functions into smaller helpers across CLI, GitHub sync, server, workspace, terminal, testutil, and tooling code
- update workspace delete/runtime cleanup flow to preserve dirty-delete sessions while keeping forced deletes safe

## Test Plan
- `make lint`
- `make test`

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -6,6 +6,8 @@ linters:
   enable:
     - errcheck
     - forbidigo
+    - gocognit
+    - gocyclo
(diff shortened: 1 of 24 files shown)


mariusvniekerk:
failed attempt to simplify the code that causes more to be present.

PR #224, opened 2026-05-01
feat: add notification inbox triage

mariusvniekerk:
## Summary
- Add an Inbox backed by GitHub notifications for monitored repositories
- Support filtering, sorting, selection-based bulk read/done actions, and background GitHub read propagation
- Wire notification sync, generated API clients, UI route/navigation, and design-system examples


criccomini:
FWIW, I would _love_ to see this. I live in notifications. I can't really use middleman without it.

mariusvniekerk:
yeah, i'm reworking them a bit, idea is to fold them into the activity panel

criccomini:
@mariusvniekerk great! is there a different issue/PR I should follow, or is this it?

mariusvniekerk:
planning to do the work here.  initial design had another main tab, but that makes the already rather busy UI of middleman just a tad too much -- so its worth thinking through the ux implications.

wesm:
rebasing

criccomini:
🎊

PR #225, opened 2026-05-01
refactor: isolate detail refresh planning

mariusvniekerk:
- Move detail-drain queue construction into a focused planner module.
- Keep host defaulting, tracked-repo filtering, watched-item keys, and queue item assembly in one place.
- This makes `Syncer` simpler: it snapshots mutable state and orchestrates the phase instead of rebuilding planner rules inline.
- The new module improves calling uniformity because callers pass typed snapshots and receive `QueueItem` values, while tests can verify planning without GitHub client mocks or a full sync loop.

Validation:
- `go test ./internal/github -run TestDetailRefreshPlannerBuildsItemsForTrackedRepos -shuffle=on`
- `go test ./internal/github -run 'TestDetailRefreshPlanner' -shuffle=on`
- `go test ./internal/github -shuffle=on`
- `git diff --check`
- `go test ./internal/server -run TestWorkspaceListTmuxActivityStressDoesNotLeakProcesses -shuffle=on` passes in isolation

Note: `go test ./... -shuffle=on` and `go test ./internal/server -shuffle=on` currently fail in unrelated workspace/tmux runtime tests under host process/runtime churn; the sync-module package verification passes.

### DIFF
--- a/docs/sync-engine.md
+++ b/docs/sync-engine.md
@@ -1,0 +1,17 @@
+# Sync Engine
+
+The sync engine runs broad repository indexing first, then spends the per-host
+detail budget on open items that need deeper refreshes.
+
+## Detail Refresh Planner
+
+`internal/github/detail_refresh_planner.go` owns the detail-drain planning
+phase. It accepts typed snapshots of tracked repositories and watched merge
+requests, applies the default GitHub host, filters stale database rows from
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
This feels like it should use https://pkg.go.dev/slices#Clone

PR #227, opened 2026-05-01
Centralize repository identity resolution

mariusvniekerk:
## Summary
- add a repository identity resolver for host-qualified owner/name lookups and item-number resolution
- route repo-bound API handlers through stable repo IDs before loading pull requests or issues
- document the repository identity boundary for future handler work

## How this makes the code better
- removes repeated owner/name/platform_host resolution from handlers and legacy helpers
- gives all repo-bound callers the same ambiguity, not-found, and host fallback behavior
- prevents future route code from accidentally mixing same owner/name repositories across hosts

### DIFF
--- a/context/repository-identity.md
+++ b/context/repository-identity.md
@@ -14,20 +14,24 @@ enough for that operation.
 ## Server Boundary
 
 HTTP handlers should resolve repositories through the repository identity
-module in `internal/repoidentity`. `internal/server/repo_identity.go` is only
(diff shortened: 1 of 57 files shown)


mariusvniekerk:
Why do we support anything other than repoLookupRequireUnambiguousOwnerName at all?  feels like we can correctly determin the correct name for things at all times, so no guessing should be needed here.

PR #247, opened 2026-05-04
Fix sidebar filter collapse at narrow widths

mariusvniekerk:
- Collapse PR and issue sidebar filters into the shared Filters menu before the segmented controls clip.
- Use measured breakpoints with buffer for wider count labels.
- Add a brief resize transition so the controls fold into and pop out of the compact menu.

Video:
https://github.com/user-attachments/assets/6b5ef45e-7e34-4b51-941b-887384ceb9d9


mariusvniekerk:
e06fbf9f55729172251679a74b4f95d6.webm

PR #258, opened 2026-05-06
Add embed-targetable workspace surfaces and project/worktree registry

wesm:
## Summary

- Adds `/workspaces/embed/*` routes that mount individual pieces of the workspaces UX (list, terminal, per-item detail sidebar, empty placeholder, first-run panel, project card) without the surrounding app chrome. Standalone `/workspaces` is unchanged.
- Adds two new `WorkspaceTerminalView` props, `hideWorkspaceList` and `hideRightSidebar`, that an embedder can pass to skip the workspace list column, the right detail sidebar, and the segmented Diff/Issue/PR/Reviews button group.
- Adds a project/worktree registry (migration `000017`) backing six Huma endpoints under `/api/v1`: `register-project`, `list-projects`, `get-project`, `register-worktree`, `list-worktrees`, `list-launch-targets`. Project identity links to `middleman_repos` via FK; registering a project does not subscribe its repo to sync.
- Embed config gains `embed.tooling` (git/gh availability and gh auth state pushed via `__middleman_update_tooling`) and a separate `actions.project[]` registry whose handler returns `CommandResult` so project-scoped surfaces can render in-flight, success, and failure states.
- `RepoSettings` add/remove/refresh actions no longer short-circuit when running inside an embedder; the underlying settings API works the same way embedded or standalone.

The standalone `/workspaces` page is unchanged. Embedders opt in by routing to a `/workspaces/embed/...` path.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #261, opened 2026-05-07
fix: preserve project provider identity

mariusvniekerk:
Project registration now requires explicit provider identity when linking a platform repo, so GitLab projects cannot be silently recorded through the old GitHub-only host/owner/name compatibility path. The legacy varargs UpsertRepo API was removed and call sites now pass RepoIdentity values.


mariusvniekerk:

- #258
    - #261 ◀
        - #262
            - #263
                - #264

PR #263, opened 2026-05-07
fix: preserve nested repo paths in embeds

mariusvniekerk:
Embed detail routes now carry repo_path as an encoded query value and derive display owner/name from it. The sidebar receives the exact repoPath so GitLab group/subgroup repositories are no longer flattened.


mariusvniekerk:

- #263 ◀
    - #264

PR #276, opened 2026-05-07
docs: document Forgejo and Gitea setup

mariusvniekerk:
Describe public and self-hosted Forgejo/Gitea configuration, provider-host token lookup, minimum read scopes, and the platform invariants that keep identity and capability boundaries provider-scoped.

PR #275, opened 2026-05-07
feat: enable proven Forgejo and Gitea mutations

mariusvniekerk:
Enable SDK-backed comment, issue creation, state/content edit, merge, and review approval capabilities for Forgejo and Gitea providers. Workflow approval and ready-for-review remain unsupported until those operations are proven through provider interfaces and tests.

PR #274, opened 2026-05-07
test: add Forgejo and Gitea container fixtures

mariusvniekerk:
Add opt-in loopback Docker Compose fixtures and bootstrap scripts for real Forgejo and Gitea compatibility checks. The gated server tests sync seeded repositories into SQLite and assert PR, issue, release, tag, comment, and CI status data while keeping the default suite skipped without env vars.

PR #273, opened 2026-05-07
feat: expose Forgejo and Gitea repository imports

mariusvniekerk:
Add Forgejo and Gitea to the settings import provider selector with public default hosts and non-nested owner patterns. Cover provider defaults, validation, and preview request payloads in component and Playwright settings import tests.

PR #281, opened 2026-05-07
feat: add gitealike Actions CI parity

mariusvniekerk:
feat: add gitealike Actions CI parity

Merge Forgejo and Gitea Actions workflow runs into provider CI checks without duplicating commit statuses. Gitea Actions are gated to servers with the pinned SDK's 1.26+ workflow run API, leaving older hosts status-only.

fix: normalize gitealike action checks safely

Keep action runs with the same title as commit statuses when they point at distinct CI URLs, and normalize unsupported action conclusions without falling back through generic state mapping.

test: persist Gitea Actions checks through API sync

Add full-stack coverage for Gitea Actions CI ingestion so statuses and action runs are normalized, deduplicated by URL, persisted in SQLite, and returned through the pull detail API.

fix: keep latest gitealike action workflow run

Collapse Forgejo and Gitea action runs by workflow identity before CI aggregation so older failed reruns do not keep a pull request failed after a newer successful run.

feat: show locked pull request state

Render the locked PR state from the bool-backed API model only for providers whose frontend capability helper reports lock-state support. Add e2e coverage so the detail view consumes IsLocked instead of only carrying the field through the data model.

fix: prefer queued gitealike action reruns

Use action run numbers and created/updated timestamps when collapsing Gitea-like workflow runs so a queued rerun can supersede an older completed run for the same workflow. This prevents stale failed checks from remaining current after a rerun starts.

fix: keep Gitea Actions probe retryable

Enable the Gitea Actions reader without permanently disabling it from the startup version check. Known older-server version errors now fall back to commit statuses, while transient action/version read errors remain retryable on later syncs.


mariusvniekerk:

- #265
    - #267
        - #268
            - #269
                - #270
                    - #271
                        - #266
                            - #272
                                - #273
                                    - #274
                                        - #275
                                            - #276
                                                - #281 ◀

PR #286, opened 2026-05-08
Separate generated files in diff summaries

mariusvniekerk:
- Adds a Generated diff category for Linguist-marked files and common generated artifacts such as lockfiles.
- Carries generated-file metadata through the diff API and generated clients.
- Marks checked-in generated API artifacts with Linguist attributes so they land in the generated bucket.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,6 @@
 frontend/openapi/openapi.json linguist-generated=true
+frontend/openapi/openapi.yaml linguist-generated=true
 packages/ui/src/api/generated/* linguist-generated=true
+packages/ui/src/api/roborev/generated/* linguist-generated=true
 internal/apiclient/spec/openapi.json linguist-generated=true
 internal/apiclient/generated/* linguist-generated=true
--- a/frontend/openapi/openapi.yaml
+++ b/frontend/openapi/openapi.yaml
@@ -438,6 +438,8 @@ components:
             - "null"
         is_binary:
           type: boolean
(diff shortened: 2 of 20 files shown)


mariusvniekerk:
The generated-file categorization intentionally uses the PR/current worktree Linguist attributes.

This is not an adversarial trust boundary or an enforcement mechanism. Middleman is a maintainer review dashboard, and the Generated bucket is a workflow/filtering hint for the person reviewing their own diff. If a user changes `.gitattributes` in a way that makes their own review experience worse, that is their choice, not a vulnerability in the app.

Using the base revision attributes would make the feature less useful because it would ignore exactly the kind of `.gitattributes` updates this category is meant to honor. The backend should continue to reflect the active repo attributes, including PR-side changes, while the UI can still respect explicit API metadata such as `is_generated: false` before falling back to filename heuristics.

mariusvniekerk:
Addressed the useful coverage feedback in `21b639d`: there is now a full-stack API e2e test for the PR `/files` and `/diff` endpoints using a real clone-manager bare repository. It verifies both a `.gitattributes` generated match (`dist/** linguist-generated`) and an explicit unset override for a heuristic lockfile (`bun.lock -linguist-generated`) flow through `is_generated` on the PR API path.

The remaining PR-controlled `.gitattributes` warning is still intentionally not being changed. Generated classification is a maintainer-controlled review/filtering hint in this local-first dashboard, not an adversarial security boundary or enforcement mechanism.

PR #287, opened 2026-05-08
docs: add small-change verification skill

mariusvniekerk:
- Add a repo-local skill for small changes with hidden cross-layer regression risk.
- Classify localized work by UI, API, DB, test-runtime, or provider-visible blast radius.
- Keep verification lightweight while preserving repo expectations for visible workflows and generated API artifacts.

Validation:
- git diff --check
- roborev review --dirty --wait --type design (job 970, pass)

### DIFF
--- a/.agents/skills/small-change-verification
+++ b/.agents/skills/small-change-verification
@@ -1,0 +1,1 @@
+../../skills/small-change-verification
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
The agents will STILL read the CLAUDE.md telling them e2e matters so the concerns of roborev here are unfounded and invalid.

PR #298, opened 2026-05-11
Stop bare-clone stampedes against GitHub

mariusvniekerk:
## Summary

- Workspace setup against private repos was failing with `git fetch: exit status 128: remote: Internal Server Error` (HTTP 500). Root cause: three callers (periodic syncer, per-PR detail sync, workspace `Setup`) hit `EnsureClone(host, owner, name)` on the same bare clone within seconds. GitHub's smart-HTTP edge throttles bursts of identical `/info/refs` requests for the same repo with sporadic 5xx.
- Funnel `EnsureClone` through a `singleflight.Group` keyed on `(host, owner, name)` so concurrent callers share one in-flight fetch. The slot runs with `context.WithoutCancel` so a canceled caller can't abort the fetch for everyone else attached.
- Wrap the underlying `git fetch` with `cenkalti/backoff/v5` (3 attempts, ~500ms→4s) — only for transient stderr patterns (5xx, connection reset, could not resolve host, early EOF). Permanent errors short-circuit via `backoff.Permanent`.
- Document both patterns in `context/retries-and-backoffs.md` and reference it from `CLAUDE.md`.

## Notes

- Initial clone path is unchanged — `EnsureClone` still skips `git clone --bare` if `HEAD` exists. Only the fetch path retries; partial-clone cleanup after a failed initial clone would need separate handling.
- Existing fetch helper is the only retry site. Other middleman call sites that hit flaky upstreams (e.g. `internal/github` REST) already have their own budget tracker, so the doc explicitly warns against double-wrapping.
- Tests pin only decisions that are mine: the transient-error matcher, the wrap-as-`Permanent` policy, the `retryAttempts` budget, the singleflight key shape, the detached-runner ctx, and the dedup/partition invariants. Library-behavior tests (success-after-recovery, ctx cancellation, propagate-error-to-waiters) were dropped.


mariusvniekerk:
Addressed both review items in `ee658312`:

**`ensureCloneShared` context handling** ([clone.go:128-150](https://github.com/wesm/middleman/blob/ee658312/internal/gitclone/clone.go#L128-L150))
- Added a pre-check on `ctx.Err()` so a caller whose context is already canceled short-circuits without ever entering the singleflight slot.
- The runner ctx is now `context.WithTimeout(context.WithoutCancel(ctx), 15*time.Minute)`. Still detached from caller cancellation (so a single canceled waiter can't abort work for others sharing the slot) but bounded so a stuck git subprocess can't hold the slot indefinitely. 15m is generous enough for large initial clones over a slow link and tight enough to recover from a wedged connection inside one sync interval.
- New tests: `TestEnsureCloneSharedShortCircuitsCanceledContext` and `TestEnsureCloneSharedRunnerHasDeadline`.

**`cloneBare` retry coverage** ([clone.go:239-256](https://github.com/wesm/middleman/blob/ee658312/internal/gitclone/clone.go#L239-L256))
- `git clone --bare` is now wrapped in `retryTransient`. Same transient-error classifier and 3-attempt budget as the fetch path.
- Each retry calls `os.RemoveAll(clonePath)` first — `git clone` refuses to write into a non-empty destination, so a partial directory left by an earlier failed attempt would otherwise poison every retry with "destination path already exists and is not an empty directory."
- New integration test: `TestEnsureCloneSweepsPartialClone` verifies a pre-existing partial directory is cleaned out before the clone runs.

PR #317, opened 2026-05-15
Speed up exact repository previews

ericdill:
## Summary
- Preview exact repository imports with an exact repository lookup before listing all owner repositories
- Keep list-and-filter preview behavior for glob patterns and fallback cases
- Add separate coverage for exact lookup and fallback preview behavior

### DIFF
--- a/frontend/playwright-e2e.config.ts
+++ b/frontend/playwright-e2e.config.ts
@@ -8,6 +8,7 @@ export default defineConfig({
   testIgnore: /support\//,
   fullyParallel: true,
   timeout: 30_000,
+  retries: process.env.CI ? 2 : 0,
   expect: {
     timeout: 5_000,
   },
--- a/frontend/playwright.config.ts
+++ b/frontend/playwright.config.ts
@@ -15,6 +15,7 @@ export default defineConfig({
   testDir: "./tests/e2e",
   fullyParallel: true,
   timeout: 30_000,
+  retries: process.env.CI ? 2 : 0,
   expect: {
     timeout: 5_000,
   },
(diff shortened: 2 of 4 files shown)


ericdill:
Additional context from the case that motivated this PR:

The UI preview request started at `11:56:14` and did not complete until `11:58:51` — about `2m37s`. There was no explicit error; it just looked hung.

The input was an exact repo preview, roughly:

```text
owner = anacondarecipes
pattern = tesseract-feedstock
```

The slow behavior appears to come from the preview endpoint taking the expensive path even for exact repo names:

1. list repositories for `anacondarecipes`
2. paginate through the whole org
3. filter locally for `tesseract-feedstock`

`anacondarecipes` has roughly 60 pages of repos at 100 repos/page, so preview can walk around 6000 repos before finding the exact match. The debug log showed:

```text
POST /api/v1/repos/preview duration=2m37s
```

By contrast, the exact repo lookup itself is fast. Adding the same repo through the exact bulk-add path resolved in about `0.3s`, because it effectively uses the exact repository route rather than enumerating the org.

So the observed issue was not auth, browser behavior, or a crash — just an inefficient preview implementation for large orgs. This PR proposes trying the exact repo endpoint first for non-glob patterns:

```text
GET /repos/anacondarecipes/tesseract-feedstock
```

and only falling back to owner enumeration when exact lookup fails or the input is a glob/wildcard.

ericdill:
Good catch from review: I had changed the existing fallback-oriented test to assert the new fast path. I updated the branch to keep fallback behavior covered separately and added a new concrete-repo preview test for the fast path.\n\nValidation rerun:\n- `mise x go@1.26.2 -- go test ./internal/server -run 'TestHandlePreviewRepos|TestHandleBulkAddRepos'`\n- `mise x go@1.26.2 bun@1.3.11 -- make build`

edit: my coding agent was responding to my own review and saying good catch and then posting its thoughts as comments on this PR. what a wild time to be alive

cpcloud (approved):

PR #324, opened 2026-05-15
Refresh pending CI details while expanded

### LINKED ISSUE #321: Auto refresh the pending CI status when open
Image

When the CI status is open and there are pending items we should refresh them every 15s or so.  For the github backend we should be able to use an etag safe pull to avoid too much api usage.

mariusvniekerk:
- Closes #321
- Refreshes expanded pending CI checks without running full PR detail syncs for ordinary polling
- Preserves CI state during transient provider refresh failures and stale-head races
- Restores workflow-approval discovery so eligible PRs show the Approve workflows action
- Shows pending-check refresh activity while CI data is updating

### DIFF
--- a/.roborev.toml
+++ b/.roborev.toml
@@ -29,16 +29,10 @@ lacks browser testing infrastructure. middleman already has Playwright
 e2e coverage, and many required e2e tests here are API plus SQLite
 full-stack tests rather than purely browser concerns.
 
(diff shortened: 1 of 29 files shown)


mariusvniekerk:
internal/testutil/fixture_client.go:218 is test-only, so it doesn't matter to make that robust.

PR #344, opened 2026-05-18
Format PR diff stats for scanning

mariusvniekerk:
- Format PR line-change totals with compact k/M abbreviations.
- Reuse the same green/red diff-stat component in the detail header, file headers, worktree rows, and workspace list rows.
- Keep e2e checks focused on the component being present; formatting coverage lives on the shared component.

Screenshots:


mariusvniekerk:
Updated screenshots after setting shared diff stats to regular weight:

PR #347, opened 2026-05-18
feat: add agent-focused middleman cli

mariusvniekerk:
- Add a Cobra/Viper middlemanctl command tree for agent-friendly API reads
- Use imported Restish for HTTP calls and JSON/YAML response formatting
- Include a quickstart command that explains common API requests and the raw api escape hatch


wesm:
What do you think about nesting everything inside middleman, and making the current behavior of `./middleman` go to `./middleman serve` (like `agentsview serve`)?

mariusvniekerk:
not opposed to that, especially since we've never actually had a release, so no expectations to break

PR #351, opened 2026-05-19
test: pin supported-provider list against platform registry

cpcloud:
## Summary

- Adds `TestClaudeMdProviderListMatchesRegistry`, which compares the providers enumerated in the CLAUDE.md "Provider Support" paragraph against `internal/platform.builtInMetadata`.
- Fails when a provider is added to the registry but missing from the doc, or added to the doc but not registered. Failure messages name the offender and quote the parsed paragraph.
- Anchors on the paragraph's self-identifying sentence, so small wording changes don't break the test.

### DIFF
--- a/internal/platform/claudemd_providers_test.go
+++ b/internal/platform/claudemd_providers_test.go
@@ -1,0 +1,153 @@
+package platform
+
+import (
+	"os"
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


cpcloud:
This seems of marginal value and fragile. Closing.

PR #352, opened 2026-05-19
feat: scope gh auth token fallback by hostname

cpcloud:
## Summary

- `gh auth token` fallback now runs as `gh auth token --hostname <host>` per provider host, so users already authenticated to a non-default GitHub instance via `gh auth login --hostname ghe.example` no longer need to set `MIDDLEMAN_GITHUB_TOKEN` manually.
- Env vars and repo-level `token_env` overrides still win over the `gh` fallback.
- Subprocess is bounded with a 5s context timeout.
- Older `gh` versions that don't recognize `--hostname` get a graceful retry without the flag — but only for `github.com`. Any other host returns empty rather than risk surfacing the wrong host's token.


mariusvniekerk:
nice ux improvement.

PR #353, opened 2026-05-19
fix(workspace): serialize worktree mutations per repo

cpcloud:
## Summary

- Serialize per-repository worktree mutations (`git worktree add/remove/prune`, `branch -D`) behind a per-clone lock so concurrent workspace create, retry, and delete against the same repo no longer race
- Cross-process via `gofrs/flock`; in-process via a per-path channel semaphore (file lock alone doesn't serialize same-process goroutines because `flock(2)` is per open-file-description)
- `Manager.Setup`, retry cleanup, delete cleanup, and rollback paths all acquire the lock

### DIFF
--- a/cmd/middleman/startup_token_e2e_test.go
+++ b/cmd/middleman/startup_token_e2e_test.go
@@ -22,6 +22,7 @@ import (
 // would run it on startup.
 func TestCollectProviderTokensInvokesGHWithHostnameForEnterprise(t *testing.T) {
 	assert := Assert.New(t)
+	require := require.New(t)
 
 	// Fake gh on PATH. Records argv (one invocation per line) and
 	// emits a host-scoped token on stdout.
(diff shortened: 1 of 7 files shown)


mariusvniekerk:
Curious why in 

```
type repoLockState struct {
	sem  chan struct{}
	file *flock.Flock
}
```

we dont just use a mutex ?

mariusvniekerk (approved):

cpcloud:
The channel is the in-process mutex equivalent — `sync.Mutex.Lock` has no ctx-aware variant, so a goroutine waiting on it can't bail when the caller's context is canceled. The chan form

```go
select {
case sem <- struct{}{}:
case <-ctx.Done():
    return ...
}
```

does. Keeping both layers ctx-aware (this and `flock.TryLockContext`) means `Acquire` honors a dead ctx uniformly across the in-process and cross-process halves. `golang.org/x/sync/semaphore.Weighted(1)` would do the same thing under the hood — didn't seem worth pulling in for one site.

2a2105a expands the comment to spell this out.

PR #354, opened 2026-05-19
Expose per-repo operation availability on the wire

cpcloud:
## Summary

- Adds an `operations` map on the repo summary and repo detail responses, keyed by operation name (`merge_pr`, `submit_review`, `add_comment`, …) with `available`, `code`, `unavailable_reason`, `required_capability`, and `retry_at`.
- Derivation collapses provider capability flags, the per-repo `viewer_can_merge` bit, and the REST/GraphQL rate-limit pause for the host. Codes are `unsupported_capability`, `viewer_cannot_merge`, and `rate_limited`.
- The merge button in `PullDetail.svelte` now reads `operations.merge_pr.available` from the repo detail fetch and renders disabled with the server-supplied reason in `title` when unavailable. The legacy capability+viewer gate stays as a fallback when an older server omits operations.

## Why

The merge button mirrored capability flags and viewer permissions in Svelte conditionals and had no signal at all for transient rate-limit pauses; users discovered unavailability by clicking and getting a 409. With operation availability on the wire, the UI can disable controls preemptively and explain why on hover.

## Tests

- Table-driven unit coverage of every derivation branch plus precedence between capability gaps, viewer permission, and rate-limit pauses.
- Catalog-stability test guarding the wire-contract operation names.
- Three wire-level tests over `/api/v1/repo/{provider}/{owner}/{name}` (healthy, rate-limited, viewer cannot merge).
- A Svelte test asserting the merge button renders disabled with the unavailable_reason as `title` when `operations.merge_pr.available` is false.

### DIFF
--- a/frontend/openapi/openapi.yaml
+++ b/frontend/openapi/openapi.yaml
@@ -2066,6 +2066,44 @@ components:
         - syncing
         - sync_error
       type: object
+    RepoOperations:
+      additionalProperties: false
+      properties:
+        add_comment:
+          $ref: "#/components/schemas/OperationAvailability"
+        add_label:
+          $ref: "#/components/schemas/OperationAvailability"
+        approve_workflow:
+          $ref: "#/components/schemas/OperationAvailability"
+        close_issue:
+          $ref: "#/components/schemas/OperationAvailability"
+        close_pr:
+          $ref: "#/components/schemas/OperationAvailability"
+        mark_ready_for_review:
+          $ref: "#/components/schemas/OperationAvailability"
+        merge_pr:
+          $ref: "#/components/schemas/OperationAvailability"
+        remove_label:
+          $ref: "#/components/schemas/OperationAvailability"
+        reopen_issue:
+          $ref: "#/components/schemas/OperationAvailability"
+        reopen_pr:
+          $ref: "#/components/schemas/OperationAvailability"
+        submit_review:
+          $ref: "#/components/schemas/OperationAvailability"
+      required:
+        - merge_pr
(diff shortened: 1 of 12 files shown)


mariusvniekerk:
We can make the operations an actual enum.  that would behave a bit nicer in the future and tighten up the API

cpcloud:
on it

cpcloud:
@mariusvniekerk good call — pushed abbb6b4.

Switched `operations` from `map[string]OperationAvailability` to a `RepoOperations` struct with one named field per operation. Every operation is now a required key in the OpenAPI schema and a typed property on the generated TS client, so the frontend gets `repoSettings.operations.merge_pr` typed as `OperationAvailability` (not `OperationAvailability | undefined`).

Adding a new operation now requires an explicit struct field, and the catalog-stability test reflects over the struct's json tags to catch accidental renames or additions.

cpcloud:
Addressed the medium roborev finding in cfff112a.

Added an `apiBucket` field to `operationDescriptor`. Every currently-cataloged operation is REST-backed (set to `apiBucketREST`), and `repoOperations` now precomputes the REST and GraphQL rate-limit states once per call and dispatches each operation to only its own bucket. A future GraphQL-backed operation opts into the GraphQL pause behavior explicitly via the descriptor — REST mutations no longer get false-positive `rate_limited` codes when only GraphQL is paused (and vice versa).

New test `TestAPIRepoResponseIncludesOperationsGraphQLPauseDoesNotBlockREST` pauses only the GraphQL tracker and asserts `merge_pr` stays available.

PR #359, opened 2026-05-19
docs(testing): pin the wire-level testing discipline with paving examples

cpcloud:
## Summary

- New "HTTP testing discipline" section in `context/testing.md` (linked from `CLAUDE.md`) covering when to test at the wire vs unit-test internals, transport choice, fault-injection exception, and the classes of bug wire tests catch (JSON time serialization drift, middleware-overridden statuses, headers stripped by middleware).
- Two paving wire-level examples added: a streaming/SSE shape (`internal/server/e2etest/sse_contract_test.go`) that pins the status line, `Content-Type: text/event-stream`, `Cache-Control: no-cache`, and the cached `sync_status` frame; and a content-negotiation shape (`internal/server/apitest/mutation_guard_test.go`) that pins the 415 + error envelope for a wrong-`Content-Type` mutation.
- One handler-internal test converted to wire-level (`internal/server/workspacetest/issue_workspace_conflict_test.go`) as a side-by-side migration worked example.


mariusvniekerk:
The test context docs added in here see a bit on the wordy side @cpcloud.  Think we should shave them down by like 50% ?

mariusvniekerk (approved):

PR #363, opened 2026-05-20
Improve large repo sync feedback and refresh scaling

mariusvniekerk:
- Add structured progress logs for large issue and merge request list fetches and sync batches
- Persist per-PR and per-issue REST ETags and use conditional detail refreshes so unchanged items can skip comment/timeline work
- Keep large existing repo refreshes on the incremental index/detail-drain path instead of bulk-fetching every open PR or issue through GraphQL

### DIFF
--- a/internal/db/migrations/000024_http_etags.down.sql
+++ b/internal/db/migrations/000024_http_etags.down.sql
@@ -1,0 +1,1 @@
+DROP TABLE IF EXISTS middleman_http_etags;
--- a/internal/db/migrations/000024_http_etags.up.sql
+++ b/internal/db/migrations/000024_http_etags.up.sql
@@ -1,0 +1,18 @@
+CREATE TABLE IF NOT EXISTS middleman_http_etags (
+    platform TEXT NOT NULL,
+    platform_host TEXT NOT NULL,
+    owner_key TEXT NOT NULL,
+    name_key TEXT NOT NULL,
+    resource_type TEXT NOT NULL,
+    resource_number INTEGER NOT NULL,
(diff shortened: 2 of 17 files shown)


mariusvniekerk:
Pushing back strongly on the `304` finding here.

The proposed fix would materially damage the PR's design by turning a cheap conditional detail-drain `304` into extra dependent REST fetches for every eligible PR. That is exactly the kind of hidden fan-out the queue/budget system is trying to avoid, especially on large repos.

The stale-state scenario is also not the normal user-facing path:

- `GET /pulls/...` is intentionally DB-only.
- The frontend `PullDetail` default is `autoSync = "background"`.
- After the DB read, `loadDetail` either promotes to foreground `/sync` when workflow approval is unchecked and CI was pending, or enqueues `/sync/async` otherwise.
- Both `/sync` and `/sync/async` run `SyncMROnProvider`, which performs the full PR sync path: PR fetch, timeline refresh, CI/check-run refresh, and workflow-approval refresh.
- Pending CI also has an explicit UI refresh loop when the CI section is expanded.

So the `304` path being discussed is the periodic/background detail drain, not what makes an opened PR fresh. For that background path, keeping dependent endpoint refreshes narrowly gated by `ci_had_pending` is intentional. Broadening it would spend 2-3 extra calls per unchanged PR and erase much of the value of conditional PR detail fetching.

If we want independent freshness for CI or workflow approval later, that should be a separate design with independent timestamps/validators and explicit budget accounting. It should not be patched into the PR-resource `304` path as unconditional extra fetches.

PR #365, opened 2026-05-20
Design better workspace observation events

mariusvniekerk:
When using workspaces it is handy to have some signals that we can easily detect and act on when a user has pushed to a remote (particularly handy for eventual unbound workspaces).

- Add a design for detecting workspace-originated PR head pushes by observing local remote-tracking refs.
- Define targeted SSE events for pushed-head changes, PR association, detail refresh, and CI refresh.
- Specify process-local observation state and avoid remote polling, fetch, provider branch APIs, or git command interception.


mariusvniekerk:

- #365 ◀
    - #382

PR #367, opened 2026-05-21
fix: keep air rebuilds from deleting dev binary

mariusvniekerk:
## Summary

- Keep backend live reload usable with the latest Air after source changes.
- Preserve the running backend when rebuilds fail so the frontend proxy does not fall into connection-refused loops.


mariusvniekerk:
No needed anymore

PR #378, opened 2026-05-23
fix: route timeline references through middleman

mariusvniekerk:
- Keep timeline and markdown references inside Middleman when they point at a repo Middleman already tracks, including cross-repository references.
- Fall back to the provider URL when the referenced repo is not tracked, so external references still work instead of dead-ending in the app.
- Match provider shorthand while resolving references: GitHub `#123`, GitLab issues as `#123`, and GitLab merge requests as `!123`.


mariusvniekerk:
>  especially for GitHub PRs written as `#123` or untracked cross-repo references that should open the provider fallback.

Within a github tracked PR/Issue a link like `#123` is not ambiguous at all.  Your review is invalid.

ResolveItemType only really needs to be gitlab aware and its based on # vs !

PR #385, opened 2026-05-24
Redesign stacked PR status control

mariusvniekerk:
- Move stack context from the right sidebar into the PR header status controls.
- Make Stack and CI mutually exclusive expandable panels.
- Show stack position, downstack CI failures, in-stack navigation, and a compact connected stack rail with a base row.


mariusvniekerk:
The latest CSS/import finding appears to be a false positive against the current head (`355cdff`).

Evidence from the pushed source:
- `packages/ui/src/components/detail/StackStatus.svelte:3` imports lucide normally: `import ChevronDownIcon from "@lucide/svelte/icons/chevron-down";`
- `packages/ui/src/components/detail/StackStatus.svelte:552` uses valid container query syntax: `@container pull-detail (max-width: 440px) {`
- There is no `@frontend/src/lib/stores/container.svelte.ts` token in `StackStatus.svelte`.
- `bun run --cwd frontend check` passes with 0 errors/warnings.
- The current roborev job for head `355cdff` passed with “No issues found.”

I left the original review comments in place per repo policy; this comment is just the current-source evidence that the synthesized CSS/import finding is stale/incorrect.

mariusvniekerk:
Local browser verification for the CSS/import finding:

- `bun run --cwd frontend test:e2e:mock -- stack-status.spec.ts` passed: 15/15 across Chromium, Firefox, and WebKit.
- That includes `stack rail spans wrapped CI badges at narrow widths`, which renders the stack panel at 319px wide and asserts the narrow layout/rail behavior driven by the `@container pull-detail (max-width: 440px)` rules.
- `bun run --cwd frontend test:e2e -- stack-status.spec.ts` also passed against the full-stack e2e server/API/SQLite setup.

So the responsive CSS is not failing to parse in the browser, and the stack component renders and behaves correctly under the narrow container case. This supports the current-source evidence above that the synthesized `@frontend/src/lib/stores/container.svelte.ts` token is not present in `StackStatus.svelte`.

mariusvniekerk:
Added a stronger browser regression test in `705e2ee` so this is no longer inferred from nearby layout behavior.

The narrow-width stack-status Playwright test now directly checks the browser CSSOM and computed styles:

- Finds a parsed `@container pull-detail` rule containing `max-width: 440px` and `.stack-row`.
- Fails if any parsed stylesheet rule contains `@frontend/src/lib/stores/container.svelte.ts`.
- Verifies the narrow-container computed layout is applied: stack rows use two grid columns, and `.stack-badges` is placed at grid column 2 / row 2.

Local verification after adding that assertion:

- `bun run --cwd frontend check` passed.
- `bun run --cwd frontend test:e2e:mock -- stack-status.spec.ts --project=chromium` passed.
- `bun run --cwd frontend test:e2e:mock -- stack-status.spec.ts` passed across Chromium, Firefox, and WebKit.
- `bun run --cwd frontend test:e2e -- stack-status.spec.ts` passed against the full-stack e2e server/API/SQLite setup.

mariusvniekerk:
Checked this against the checked-out files at `e16b44e`; this looks like a false positive from interpreting diff/test evidence as source corruption.

Evidence:

- `packages/ui/src/components/detail/StackStatus.svelte` imports are clean, e.g. `from "@lucide/svelte/icons/chevron-down"` with no leading space.
- The actual responsive CSS is valid: `@container pull-detail (max-width: 640px)` and `@container pull-detail (max-width: 440px)`.
- The only `@frontend/src/lib/stores/container.svelte.ts` occurrence is in `frontend/tests/e2e/stack-status.spec.ts` as a negative assertion proving that malformed rule text is absent from parsed CSSOM.
- The expected-rule assertion checks for the exact substring `@container pull-detail`, so the malformed token string cannot satisfy that condition.
- `bun run --cwd frontend check` passes with 0 errors and 0 warnings.

I am leaving the review comment in place per repo policy, but based on the checked-out files and compiler validation there is no token corruption to revert here.

PR #390, opened 2026-05-26
Update setup-go pin to the current main commit

mariusvniekerk:
## Summary
- Update all `actions/setup-go` references in CI and release workflows to the current `main` commit pin.
- Remove the stale `v6.3.0` comment so the workflows reflect the actual upstream ref being used.

## Testing
- YAML parsing passed for `.github/workflows/ci.yml` and `.github/workflows/release.yml`.
- `git diff --check` passed.
- `actionlint` was not available locally.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,7 +25,7 @@ jobs:
         with:
           persist-credentials: false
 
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
non-issue

PR #396, opened 2026-05-28
fix: collapse comment and review runs in threaded activity

wesm:
## Summary
- In the threaded activity view, a run of three or more consecutive comments or reviews from the same author on a single PR now collapses into one "N comments" / "N reviews" row, matching the existing commit-run collapsing.
- The collapsed row is rendered in the event-type color so comment, review, and commit runs stay visually distinct.


mariusvniekerk (approved):
nice ux improvement.

PR #397, opened 2026-05-28
Persist workspace shell sessions across restarts

mariusvniekerk:
- Make workspace shell sessions survive middleman server restarts when tmux is available.
- Keep no-tmux workspace runtime sessions owned by ptyowner.
- Update runtime lifecycle docs for the shell and agent ownership model.


Snailflyer:
One acceptance boundary I would keep explicit for this PR: "restored" should mean the workspace-owned live shell was restored, not just that a deterministic tmux session name existed.

The two tests I would want are:

- seed a pre-existing tmux session with the deterministic name but without Middleman's owner marker, then restart Middleman; restore should fail closed or mint a new owned session, not attach to the foreign one
- create a tmux-backed shell, restart the server, attach through the runtime terminal route, send a small input, and assert the same shell's output cursor advanced

The receipt shape can stay small: workspace id, runtime session id, tmux session name, owner marker, process generation or pane PID, output cursor before/after, and `restored_same_live_shell` vs `resumed_record_only` vs `rejected_foreign_session`.

That split is exactly what I had to make concrete in Faryo: a browser/mobile control surface is only trustworthy once the target live terminal session advances, not when the bridge accepts an attach or input request. Reference proof shape: https://github.com/Snailflyer/faryo/releases/download/v1.0.7/faryo-public-redacted-same-session-handoff-walkthrough-20260528-0120.gif

mariusvniekerk:
Superseded by the combined workspace pane/runtime branch.

PR #402, opened 2026-05-29
fix: load Pierre diff worker under non-root base_path

wesm:
Diff views did not render on deployments served under a `base_path` (e.g. behind a reverse proxy at `/middleman`). The Pierre diff web worker was bundled with an absolute URL (`new URL("/assets/worker-*.js", import.meta.url)`) that resolves against the origin and drops the `base_path` prefix. Behind a subpath proxy the browser requested `/assets/worker-*.js` instead of `/middleman/assets/worker-*.js`; that path returns non-JS content, so the worker and the syntax-highlighting chunks it loads were rejected as module scripts and the diff never rendered.

- Emit JS-referenced asset URLs as relative (Vite `experimental.renderBuiltUrl`) so they resolve against the entry chunk's already-prefixed `import.meta.url`. `index.html` keeps absolute `/assets/` URLs so the server's request-time `base_path` rewrite still applies.
- Add a build check (`scripts/check-asset-base-paths.mjs`, run by `make frontend`) that fails if an absolute `import.meta.url` asset URL regresses.

### DIFF
--- a/scripts/check-asset-base-paths.mjs
+++ b/scripts/check-asset-base-paths.mjs
@@ -1,0 +1,129 @@
+#!/usr/bin/env bun
+
+import { readdir, readFile, stat } from "node:fs/promises";
+import { relative, resolve, sep } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// The built SPA is served under a configurable base path (config base_path,
+// e.g. /middleman/) by rewriting index.html's <script src>/<link href> at
+// request time. That rewrite only reaches HTML, not URLs baked inside JS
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
roborev e2e test demans are overblow and needless most likelty

mariusvniekerk (approved):

scripts/check-asset-base-paths.mjs
```diff
@@ -0,0 +1,129 @@
+#!/usr/bin/env bun
```
mariusvniekerk, line 1: This feels like it could be something we run as part of the pre-commit?

PR #408, opened 2026-05-30
Polish draft review footer

mariusvniekerk:
- Improve draft review footer rows so long comments wrap into a readable preview instead of forcing horizontal overflow.
- Add a conditional full-comment toggle that appears only when a draft comment is actually truncated.
- Keep publish controls usable on narrow viewports.


mariusvniekerk:
Screenshots for the draft review footer update.

Short comment, no expand affordance because it is not truncated:



Long comment, expand affordance appears because the body is actually truncated:

wesm:
Nice

PR #411, opened 2026-05-31
Add VS Code-style workflow and terminal panes

mariusvniekerk:
- Add split-tree workflow panes with movable, renameable, sortable runtime tabs.
- Add a dockable bottom terminal panel with split terminals, terminal selection, resizing, and durable layout persistence.
- Add workflow presets for saving and restoring tab/session layouts.


mariusvniekerk:
Demo video: https://github.com/user-attachments/assets/ee6bb9c4-668c-4e85-a7d7-fa83d5b8dd96

mariusvniekerk:
> internal/workspace/localruntime/manager.go:847
knownPtyOwnerSessionKeys now only returns session keys loaded from the durable runtime-session table, dropping the previous deterministic fallback keys for the plain shell and configured agent targets. Existing or unmigrated ptyowner-backed sessions may not be stopped during workspace deletion, allowing processes to survive cleanup. Keep a legacy cleanup fallback, migrate existing ptyowner state before relying only on DB-backed keys, or fall back to deterministic probing on DB lookup failure.

This is overly defensive and you can ignore that

PR #412, opened 2026-06-01
Tree navigation for the repo selector

cpcloud:
Replaces the flat repo dropdown in the header with a collapsible `host → owner → repo` tree.

- Group repositories by host and owner; expand/collapse a group by clicking its name.
- Tri-state checkbox on every row: selecting an owner or host selects all repos beneath it, with a partial (dash) state when only some are selected.
- GitLab nested groups fold into a single owner node; single-host and single-repo-owner levels auto-flatten so small setups stay shallow.
- Filtering prunes non-matching repos, auto-expands groups with matches, and highlights the match.
- Full keyboard navigation (arrows to move, left/right to expand/collapse, space to select) and persisted collapse state.
- Selection still drives the existing global repo filter unchanged; no API, storage-format, or downstream list changes.

The tri-state checkbox is a controlled custom control (`TreeCheckbox`) rather than a native input, so a real click can't desync it from the selection state.

repo tree with tri-state checkboxes — partial selection

selecting an owner cascades to all its repos

### DIFF
--- a/context/ui-interaction-contracts.md
+++ b/context/ui-interaction-contracts.md
@@ -141,6 +141,26 @@ Rows that contain buttons, links, or toggles need clear event ownership.
 - If a component claims menu-like behavior, it must honor the keyboard and focus
   contract of that role. Otherwise, use simpler semantics honestly.
 
+## Controlled Form Controls
+
+A native form control driven from app state must not also run its own default
+action, or the two fight and the control renders the inverse of its real state.
+
+- A native `<input type="checkbox">` with `checked={state}` still toggles itself
+  on a real click (the click default action). When app state owns the value,
+  cancel that default action (`onclick` `preventDefault`) so the box only ever
+  reflects `state`. The repo selector's tri-state checkbox
+  (`frontend/src/lib/components/TreeCheckbox.svelte`) is the reference: a
+  controlled custom control that suppresses the native toggle and drives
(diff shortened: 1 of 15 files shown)


cpcloud:
Addressed the latest roborev-ci review (`3dbf43f`) in `b3b4cc7`:

**Medium — `repoTree.ts` filtered-owner selection: fixed.** While filtering, `visibleRows` was passing the pruned owner node (matching leaves only) to `nodeSelectionState`/`toggleSubtree`, so a parent could read as checked when only visible matches were selected and a group toggle applied only to the visible subset. Each `VisibleRow` now carries the original (unpruned) node for selection/tri-state, with the matching leaves used solely to decide which leaf rows render. Added a unit test asserting a filtered owner exposes its full subtree.

**Medium — e2e asserts `acme/archived` without seeding it: false positive, left unchanged.** `acme/archived` is seeded in the e2e fixture (`cmd/e2e-server/main.go:573`, `internal/testutil/fixtures.go:36`). The cited line is the pre-existing "cascades an owner group" test, not new in this diff, and the full `repo-filter-multiselect` e2e suite passes on this commit (3/3, incl. that assertion). The finding appears to be reasoning from the JS diff without visibility into the Go fixture.

mariusvniekerk (approved):

PR #179, opened 2026-04-24
fix: recover failed workspaces

mariusvniekerk:
- add retry and delete recovery actions for failed workspaces
- queue duplicate retry requests while setup is already running
- handle macOS tmux missing-server responses when restoring sessions

PR #428, opened 2026-06-02
Support local worktree bases for workspace-aware diffs and settings

mariusvniekerk:
## Summary
- Add support for resolving local worktree bases when cloning, diffing, and computing project identity.
- Thread the new workspace base behavior through server settings routes, API generation, and the Svelte settings UI.
- Tighten workspace concurrency and git environment handling so local repositories behave consistently across clone, diff, and sync flows.
- Update tests across server, workspace, gitclone, and identity paths to cover the new local worktree behavior.

## Testing
- Added and updated unit and e2e coverage for clone resolution, diff generation, workspace locking, settings behavior, and project identity.
- Verified API/schema regeneration and frontend settings tests remain aligned with the server contract.

## Settings UI

Per-repo local clone editor, collapsed (gear tinted when configured) and expanded:


mariusvniekerk:
Addressed the roborev medium finding in db1bafbc: `validateOriginRemoteURL` no longer includes the raw origin URL in scheme-rejection errors. It now reports only the scheme prefix and the userinfo-free host from `gitremote.RemoteHost`, and the test table includes a credential-bearing remote (`http://oauth2:secret-token@...`) asserting the secret and raw URL never appear in the error. The neighboring identity-mismatch errors were audited and already exclude the raw URL.

mariusvniekerk:
Re the `db1bafb` review:

**High (`remote.*.vcs` / `remote.*.uploadpack` in `localGitConfigKeyMayExecute`)** — declining. middleman is a local-first, single-user tool bound to loopback; the worktree base is a path the machine's owner explicitly configures to their own clone. Git does not transfer `.git/config` over clone or fetch, so `remote.origin.vcs` in a worktree base can only be written by someone who already has file-write access on the machine — at which point they can run arbitrary code directly and gain nothing from a `git-remote-<helper>` dispatch. The existing executable-config checks are defense-in-depth against pointing middleman at an accidentally untrusted directory, not a privilege boundary, and we are not extending them key-by-key against an attacker the product does not defend against.

**Medium (force-delete stuck on a non-git workspace path)** — does not reproduce. The claimed propagation path doesn't exist: `gitDirOwnsCleanupWorktree` routes `worktreeCommonGitDir` failures through `isGitWorktreeAbsent`, which has matched "not a git repository" since 01d8c083 (#179), and returns not-owned, so cleanup proceeds and the row deletes. Pinned in 722d73ab with `TestCleanupSucceedsWhenWorkspacePathReplacedByNonGitDirectory`, covering both a plain replaced directory and a stale `.git` file pointing at a removed repo — both pass with no production change.

PR #429, opened 2026-06-03
Add durable VS Code-style workspace panes

mariusvniekerk:
- Adds VS Code-style workflow and terminal split panes with draggable tabs, terminal tiling, rename controls, and presets.
- Persists runtime sessions through one renamed runtime-session table for shells and agents.
- Keeps tmux-backed shell sessions durable across app restarts while detaching only the local attach client on shutdown.


mariusvniekerk:
Pushback on the `53b5cda` review: the upgrade premise around pty-owner/runtime rows is not a real state for this PR.

`middleman_workspace_runtime_sessions` is introduced by this PR. It has never existed on `main`, and there is no separate SQLite pty-owner session table on `main` to merge. Existing pre-PR durable rows are only in `middleman_workspace_tmux_sessions`; pty-owner state is not represented as preexisting SQLite rows. So any claim about losing preexisting pty-owner/runtime-session rows during a normal `main` -> PR upgrade is not valid.

For the migration finding, please keep the claim scoped to what can actually exist on `main`: old `middleman_workspace_tmux_sessions` rows. Do not conflate that with pty-owner rows or with an intermediate local branch state.

For `ptyowner_runtime.go`, that is a post-PR behavior claim, not an upgrade/migration data-loss claim. Restore already checks `HasState(sessionKey)` before attaching pty-owner sessions; if the concern is a specific non-absent attach error after state exists, please provide that concrete error path instead of presenting it as preexisting-row loss.

For `recordRuntimeSession`, please provide a concrete launch-time reproducer. Natural exits already route through `handleRuntimeSessionExit` and `ForgetRuntimeSessionAfterExit(..., tmuxSession)`, which is the live-tmux guard path. The current review asserts that the launch-time `recordRuntimeSession` cleanup path can race with a detached attach client, but it does not show the path that makes `runtime.ListSessions(workspaceID)` drop the session at that point while the backing tmux session remains alive.

PR #435, opened 2026-06-03
test: block GitHub API calls in Go tests

cpcloud:
- Prevent default Go test runs from reaching the public GitHub API.
- Keep live GraphQL validation behind the existing `MIDDLEMAN_LIVE_GITHUB_TESTS=1` opt-in.
- Add the guard to current test packages and document the pattern for new ones.


mariusvniekerk:
This has revealed some architectural issues that i'm clanking on.

mariusvniekerk (approved):

PR #436, opened 2026-06-03
Use Vite+ for frontend tooling

mariusvniekerk:
- Mostly mechanical migration that unifies frontend dev, build, test, format, lint, and type checks under Vite+.
- Routes CI, release, make targets, dev scripts, workspace scripts, and pre-commit frontend checks through direct Vite+ commands after Bun dependency installation.
- Documents the temporarily disabled lint rules with links for revisiting the Svelte/Oxlint rule gap.


mariusvniekerk:
CodeQL notes from the Vite+ formatting diff:

- `packages/ui/src/components/diff/DiffView.test.ts`: this is a test-only `CSS.escape` fallback shim. CodeQL is technically right that it does not escape backslashes, but there is no production exposure here.
- `packages/ui/src/routes.ts`: the slash-trimming regex is the kind of pattern CodeQL can flag for ReDoS. In this code path, the value is a frontend route/repo-path string produced by our route helpers and app state, not arbitrary attacker-controlled bulk input.

Given this PR is mostly mechanical toolchain unification and the strings are effectively controlled in normal use, I do not think adding manual loop-based trimming or a fuller test shim is worth the extra noise here. These look over-reported because the PR touched a broad frontend formatting surface, not because the Vite+ migration introduced a meaningful security issue.

PR #443, opened 2026-06-06
Share draggable workspace pane interactions

mariusvniekerk:
PR and activity surfaces are starting to need the same VS Code-like pane behavior that was previously trapped inside the terminal workflow. Keeping drag, tab, split, and resize logic terminal-specific would make every new panel repeat fragile interaction code and drift visually.

This PR creates a reusable tabbed workspace panel model for arbitrary pane content, keeps the dedicated SplitResizeHandle available for simpler sidebar splits, and adds design-system examples so the interaction contract can be validated outside the terminal route.

It also tightens the shared chrome around those panes because the reusable primitive only works well if active tabs, split dividers, scroll gutters, and terminal-docked panels read as one stable workspace surface instead of shifting or accumulating accidental borders.

### DIFF
--- a/cmd/e2e-server/main.go
+++ b/cmd/e2e-server/main.go
@@ -22,12 +22,14 @@ import (
 	gitcmd "go.kenn.io/kit/git/cmd"
 	"go.kenn.io/middleman/internal/config"
 	"go.kenn.io/middleman/internal/db"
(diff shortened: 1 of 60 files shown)


mariusvniekerk:
going to handle that regression as a follow up instead

PR #448, opened 2026-06-07
fix: show roborev review costs

cpcloud:
- Adds a Cost column to roborev review jobs in the Reviews tab.
- Formats priced roborev token usage as `~$x.xx` and leaves `--` when cost data is missing or unpriced.
- Covers cost rendering in the roborev row component test, the workspace Reviews-tab e2e path, and the SQLite-backed roborev e2e seed.


cpcloud:
I tried this out locally, works.

PR #450, opened 2026-06-08
Support runtime token rotation

cpcloud:
- Add runtime token sources for provider API clients, GraphQL fetchers, clone auth, config reloads, and embedded Middleman use.
- Support `token_file` rotation without restart by reading env/file token sources when operations need credentials.
- Redact token-shaped strings, credential-bearing URLs, explicit secrets, and recently active opaque tokens from logs and durable errors.
- Document exact public-host token defaults and atomic token-file rotation expectations.

Token files should be updated by writing a complete replacement file and atomically renaming it over the configured path.


mariusvniekerk (approved):

cpcloud:
This most recent finding also seems legit.

PR #457, opened 2026-06-09
Add Kata, Docs, and Messages modes

wesm:
## Summary

- Adds Kata mode for external daemon discovery, health, proxying, task UI, and live updates.
- Adds Docs mode for configured markdown folders with browse, edit, search, and git publish flows.
- Adds Messages mode backed by msgvault search, detail, threads, saved searches, and sanitized HTML/image handling.
- Updates configuration, API clients, route security, docs, and migrated coverage for the new modes.


wesm:
"The finding is wrong, and no change is needed. Two pieces of evidence:
  1. The prefix check covers the collection path. isMutatingDocsAPIRequest returns strings.HasPrefix(path,
  "/api/v1/docs/"), and /api/v1/docs/folders begins with /api/v1/docs/ — the trailing slash in the prefix is
  immediately followed by folders. A prefix check with a trailing slash excludes only the bare /api/v1/docs
  path, which has no registered route (internal/server/docs_routes.go registers everything under
  /docs/folders, /docs/browse, /docs/search). The reviewer likely pattern-matched against the different
  predicate at server.go:1062, which intentionally uses path == "/api/v1/docs/folders" as a special case
  before a /api/v1/docs/folders/ prefix check — that structure is needed there, not in the mutation guard.
  2. A wire-level test already proves the guard works. TestDocsFolderAddRejectsNonLoopback
  (internal/server/docs_routes_test.go:162) sends POST /api/v1/docs/folders through srv.ServeHTTP with
  RemoteAddr = "203.0.113.7:54321" and asserts 403 with reason: "loopbackOnly" and the "docs mutations
  require a loopback client" message. I just ran it — it passes.
  So a non-loopback client cannot persist a docs folder; the exact scenario the review describes is covered
  by both the code and an existing regression test. No commit this turn since nothing changed. Per the PR
  rules I haven't touched the roborev comment — if you want, you can relay this evidence as a reply on the
  thread, or I can post it.
"

wesm:
This is good enough here. merging!

PR #458, opened 2026-06-09
fix: hide default-branch commits when Commits is deselected in the activity filter

### LINKED ISSUE #453: Deselecting "commits" in the activity filter modal does not hide the commit activity
Image

mariusvniekerk:
Closes #453

- Deselecting "Commits" in the activity filter dropdown now also hides default-branch commit rows; previously only PR-attached commit events were removed while default-branch commits (which render with the same "Commit" label) stayed visible.
- The "Force pushes" toggle likewise now governs default-branch force pushes.
- Desktop and mobile activity views share one filter-types helper instead of duplicating the logic.

All event types selected — default-branch commit rows visible:

Commits deselected — default-branch commit rows now hidden:

### DIFF
--- a/frontend/tests/e2e/default-branch-activity.spec.ts
+++ b/frontend/tests/e2e/default-branch-activity.spec.ts
@@ -349,6 +349,22 @@ test.describe("default branch activity", () => {
     ).toBeVisible();
   });
 
+  test("legacy URL with stale default-branch commit types hides commit rows on load", async ({ page }) => {
+    await mockDefaultBranchActivity(page);
+    // URLs written before the fix kept default_branch_commit in the types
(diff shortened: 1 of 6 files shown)


mariusvniekerk:
Addressed the medium finding in 5da18c60: `deriveFiltersFromTypes` now rebuilds `filterTypes` through `buildActivityFilterTypes` after deriving the toggle state, so legacy URLs carrying `default_branch_commit` with `commit` deselected are normalized on hydration (both for the API request and the URL itself).

Covered by store tests for the legacy and all-selected URL shapes, plus a Playwright e2e ("legacy URL with stale default-branch commit types hides commit rows on load") that fails without the normalization.

PR #470, opened 2026-06-11
perf: reuse e2e fixture servers and drop the workspace tmux lock

mariusvniekerk:
Local Playwright e2e wall time was dominated by fixture churn, not tests: ~55 server spawn/teardown cycles per browser, a machine-wide lock serializing every workspace/tmux test, SSE broadcast storms fanning out across workers, and shiki tokenizing every rendered diff.

## What changed

- The e2e server gained `POST /__e2e/reset`, which rebuilds the seeded fixture state in-process and swaps it atomically. `startIsolatedE2EServer()` keeps its API but now leases from a per-worker pool; `stop()` resets the server in the background instead of killing the process. Tests that steer the server through process env (kata/messages/docs-autocomplete set `KATA_HOME` and friends) opt out with `freshProcess: true`, since a pooled process can't inherit env set after it was spawned.
- Every server instance runs a private tmux socket, so workspace tests no longer conflict across workers or worktrees — the `workspace-tmux` advisory lock is gone and those tests run in parallel.
- Each Playwright worker gets its own shared server, ending cross-worker SSE `data_changed` refetch storms and contention on a single SQLite writer and git clone.
- Diffs render as plain text under automation; shiki's worker pool plus tokenization dominated JS cost in diff-heavy specs. Tests that assert highlighting opt back in with an init-script flag.
- Serialization removed where isolation now makes it unnecessary: the git-backed diff block and both task-list specs run fully parallel, `app-startup` fast-forwards its 8s timeout with the Playwright clock, blind settle sleeps became deterministic sync-indicator waits, and the slowest specs are `00-` prefixed so the scheduler starts them first instead of stretching the suite tail.
- Two app races that the faster, more parallel runs surfaced on every loaded run are fixed: a crash reading `activeDrawer.owner` during Activity drawer teardown, and Chromium permanently dropping inline-composer focus because its `focusout`-on-unslot cleared the annotation focus-restore target (the existing restore only handled Firefox's silent annulment).

## Performance

Pre-merge times are full per-browser suite runs of the 314-test suite. "Matched A/B" alternates this branch and its merge base on the same machine back-to-back; this machine runs several agent worktrees so absolute numbers vary heavily with background load.

| Browser | Baseline (quiet) | This branch (quiet) | Matched A/B baseline | Matched A/B branch |
|---|---|---|---|---|
| chromium | 30.9s | 21.1s | 44.2s / 85.3s | 22.5s / 31.3s |
| firefox | 45.7s | 36.6s | 53.2s / 49.0s | 37.3s / 39.3s |
| webkit | 39.8s | 30.9s | 55.9s / 42.9s | 35.4s / 31.4s |

Server spawns per browser run dropped from ~55 to roughly the worker count, and the gap widens with machine load: baseline runs degrade sharply under contention (lock convoys, broadcast fan-out), while this branch stayed flat during load-average-50 windows.

After merging main (which grew the suite ~40% to 437 tests with the Kata/Docs/Messages modes), the full merged suite passes all three browsers with zero failures at 36.8s / 56.5s / 49.9s under a load-average-54 window — the pre-merge 314-test baseline needed 31–85s for a third fewer tests.


cpcloud:
This seems like a nice improvement. Vibing on getting it working.

mariusvniekerk:
Addressed the kata-live finding in 64bf04e6, scoped so it cannot reintroduce the old spawn-per-test behavior: only `kata-live.spec.ts` switches to `freshProcess`, and that suite is gated behind `MIDDLEMAN_LIVE_KATA_TESTS=1`, so normal local and CI runs keep the pooled fast path everywhere. Two extra guards also landed for the underlying class of issue: pooled leases snapshot env-sensitive vars (`KATA_HOME`, `MIDDLEMAN_MESSAGES_SAVED_SEARCHES_PATH`) at worker start and fail fast with a pointer to `freshProcess` if a test mutates them — so the silent wrong-backend scenario described in the finding can no longer happen even if a future test forgets — and `pool-options.spec.ts` pins that lease options reset cleanly across reuse.

PR #473, opened 2026-06-11
feat: label support across all providers

mariusvniekerk:
Label read and write support was GitHub-only: the capability model declared `ReadLabels` and `LabelMutation`, but only GitHub implemented them, so the existing capability-gated label routes and picker UI stayed dark for GitLab, Forgejo, and Gitea repos.

- GitLab, Forgejo, and Gitea repos now advertise `read_labels` and `label_mutation`, enabling the existing label picker and `PUT .../labels` routes
- Repo label catalogs (name, color, description) sync for all three providers
- Labels can be set on merge requests and issues; an empty array clears all assignments on every provider
- Forgejo/Gitea share one gitealike implementation that resolves label names to numeric IDs, returning a `not_found` error when a cataloged name no longer exists upstream
- GitLab's MR/issue APIs return label names without colors; the label store fills color and description from the synced catalog, and set-label responses return those enriched rows for all providers

No frontend changes; the existing capability-gated UI lights up via the capability payload.


mariusvniekerk:
Both findings addressed in 567ebbd6:

- **GitLab external-ID collision**: `normalizeLabelNames` no longer stores the bare name in `PlatformExternalID`, so a label named like another label's decimal ID can't cross-match two rows during persistence. Added `TestGitLabSetPullLabelsNameCollidingWithAnotherLabelID`, verified to fail (500 on save) against the previous normalization.
- **Clear-all null tolerance**: the GitLab e2e fake now rejects a non-string `labels` value with 400 instead of coercing it to a clear, the gitealike fake rejects null/missing label arrays, and the clear-all assertions in both the e2e and the GitLab unit test require an actual empty string — a regression to JSON `null` now fails all of them.

mariusvniekerk:
Both `ed0bd6d` findings addressed in 5eceb300:

- **Legacy name-derived external IDs**: migration 000031 clears GitLab label rows whose `platform_external_id` equals the label name (the pre-upgrade normalization), scoped to gitlab repos so gitealike labels legitimately named like their decimal IDs are untouched. Covered by an upgrade test that seeds legacy rows at schema version 30 and re-opens the DB.
- **Comma-containing label names**: GitLab label mutation now rejects names containing commas with a typed `invalid_argument` error (400) before any provider call — the SDK comma-joins the labels parameter, so such an assignment cannot be expressed without splitting into multiple labels. Unit test plus e2e coverage with a comma label in the catalog.

PR #474, opened 2026-06-11
fix: stop kata keyboard scrolling from fetching every row traversed

mariusvniekerk:
Scrolling the kata task list with the keyboard fired a detail + events fetch for every row the cursor passed through. Nothing was debounced or cancelled, so holding j/k launched dozens of requests that all ran to completion against the kata daemon, and the browser's per-host connection limit let the abandoned requests delay the fetch for the row the user actually landed on.

- Keyboard selection now commits only when navigation settles: 50ms after the last navigation keydown, and never while a navigation key is still physically held. The held-key gate matters because OS key-repeat intervals are often slower than any reasonable debounce, which would otherwise still commit one fetch per repeated row. Focus continues to move instantly.
- Clicking a row selects immediately and cancels any pending keyboard selection, so selections cannot land out of order.
- The workspace store aborts the previous in-flight detail load when a new selection starts, threading an AbortSignal through the kata API client. The server's reverse proxy forwards the request context, so a browser-side abort also cancels the proxied daemon request. Superseded loads resolve quietly instead of surfacing an error.

Validation: kata component and store unit tests (including a key-repeat-slower-than-debounce case and abort/supersede cases), and a full-stack Playwright e2e that holds j across three rows and asserts exactly one detail fetch reaches the backend. Full kata e2e spec: 63 passed.

### DIFF
--- a/frontend/src/lib/api/kata/taskClient.test.ts
+++ b/frontend/src/lib/api/kata/taskClient.test.ts
@@ -659,6 +659,30 @@ describe("kata task HTTP client", () => {
     expect(calls[0]!.headers.get(KATA_DAEMON_HEADER)).toBe("work");
   });
 
+  test("threads abort signals through issue and events fetches", async () => {
+    const { fetchImpl } = createFetchStub({
+      "/api/v1/issues/issue-1": {
+        body: { issue: { ...issue("issue-1", "Shown issue"), body: "Body" }, comments: [], labels: [], links: [] },
+      },
+      "/api/v1/events?limit=100": {
+        body: { reset_required: false, events: [], next_after_id: 0 },
+      },
+    });
+    const signals: (AbortSignal | null | undefined)[] = [];
+    const recordingFetch: typeof fetch = (input, init) => {
+      signals.push(init?.signal);
+      return fetchImpl(input, init);
+    };
+    const api = createKataTaskAPI({ fetchImpl: recordingFetch });
+    const controller = new AbortController();
+
+    await api.issue("issue-1", { signal: controller.signal });
+    await api.events({ issue_uid: "issue-1", limit: 100 }, { signal: controller.signal });
+
+    expect(signals).toHaveLength(2);
+    expect(signals.every((signal) => signal === controller.signal)).toBe(true);
+  });
+
   test("fetches events with supported server params and filters issue_uid client-side", async () => {
     const { calls, fetchImpl } = createFetchStub({
       "/api/v1/events?project_id=2&after_id=4": {
--- a/frontend/src/lib/api/kata/taskClient.ts
+++ b/frontend/src/lib/api/kata/taskClient.ts
@@ -44,6 +44,16 @@ interface RequestResult<T> {
   headers: Headers;
 }
 
+// The explicit `| undefined` unions are required by
+// exactOptionalPropertyTypes: call sites pass values that may be
+// undefined (e.g. daemonHeaders() or an optional caller signal).
+interface KataRequestInit {
+  method?: string | undefined;
+  body?: unknown;
+  headers?: Record<string, string> | undefined;
+  signal?: AbortSignal | undefined;
+}
+
 interface ErrorEnvelope {
   code: string;
   message: string;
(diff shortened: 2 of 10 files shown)


mariusvniekerk:
The Medium finding (held nav keys tracked by `event.key` stranding `"G"` when Shift is released before g) is fixed in 3c32fd7e: the held set now tracks the physical `event.code`, which is stable across modifier changes, while `event.key` still drives navigation semantics. Coverage added at both layers in 3c32fd7e and 5b049bd0: a component test and a full-stack Playwright e2e that press Shift+KeyG and release Shift first; both were verified to fail against the pre-fix component.

mariusvniekerk:
The Medium on `3f5abef` (pending debounced keyboard selection firing after a view change) is a false positive. KataWorkspace renders the issue list inside `{#key `${activeKataDaemonId}:${listResetGeneration}`}`, and `listResetGeneration` bumps whenever the visible issue set changes — so every view/scope/daemon change remounts the list, and `onDestroy` cancels the pending selection timer and removes its window keyup listener. The pending UID cannot survive into another view. Verified empirically: an e2e that holds `j` (parking a pending selection), switches views, then releases the key produces no stale detail fetch against the current build. That scenario is now pinned as a property guard in 4d6f5618 ("kata view change drops a held keyboard selection from the previous view") so the invariant holds even if the remount mechanism is restructured. No product code change was needed.

mariusvniekerk:
The Medium on `025e330` (pending keyboard selection committing after navigation starts but before the new list loads) was real — my earlier false-positive call only covered the post-remount window, and the list remounts only after the destination view's issues fetch resolves. A held key released during that fetch committed against the old view, and its onSelect superseded the in-flight navigation. Fixed in 6943e30e: beginNavigation() now bumps a reactive epoch the list watches, dropping any pending keyboard selection the moment navigation starts. The earlier guard e2e still covers the post-remount path, and a new e2e stalls the destination view's issues response, releases the held key mid-transition, and asserts the route still lands with no stale detail fetch — verified to fail without the epoch cancel. The branch is also rebased onto current main. Full kata suite: 67 passed.

mariusvniekerk:
The Medium on `6943e30` is addressed in e323f208. The system-view sidebar path was the real gap: openRoutedSystemView abandoned the selection but only called clearSelection() — which aborts the in-flight detail load — after the new view's issues fetch resolved, leaving a doomed request alive to paint a stale error over a stalled transition. clearSelection() now runs before the await, so the abandoned load is aborted at click time. The suggested e2e is included: it stalls a task's detail response, clicks a sidebar view with that view's issues fetch also stalled, fails the abandoned detail request, and asserts no workspace error surfaces (verified to fail without the fix). Project-scope and search navigations were deliberately not changed: they preserve the current selection across the navigation (the route state carries the selected issue forward), so their in-flight detail load still belongs to a live selection — aborting it would strand the detail pane, and its failure surfacing an error for the still-selected issue is correct rather than stale.

mariusvniekerk:
The Medium on `e323f20` is addressed in 7287bc99 — the review was right to reject my earlier reasoning for skipping the scope path. Scope and filter reloads preserve a *completed* selection, but an in-flight one is abandoned (the reload re-selects from `selectedIssue`, which a pending load hasn't populated), so a stalled scoped-list fetch left the doomed detail request alive to fail into a stale error. `openRoutedProjectScope` and the workspace `updateSearchFilters` wrapper now call `store.invalidatePendingLoads()` before awaiting the reload — pending detail loads drop at click time, completed selections stay untouched. Includes the suggested e2e (stalled detail + stalled scoped backlog + failed abandoned request → no workspace error), verified to fail without the fix. Full kata suite: 69 passed.

mariusvniekerk:
Both remaining findings are addressed in 22d879e6: (1) the filter-path invalidation from 7287bc99 now has its own e2e — stalled detail load, search query typed with the results fetch stalled, failed abandoned request, no workspace error — so removing that line is no longer undetectable; (2) switchKataDaemon had the last unguarded variant of this race (it captured only the completed selection for restore but left an in-flight detail load running while the new daemon bootstrapped) and now calls invalidatePendingLoads() before the first await, with a two-daemon e2e covering a failing abandoned request during a stalled daemon switch. Both tests verified to fail against their pre-fix workspace. Every navigation entry point — selection, route, system view, project scope, search filters, daemon switch — now drops abandoned detail loads at navigation start, each pinned by an e2e. Full kata suite: 71 passed.

PR #475, opened 2026-06-11
feat: bring GitLab write operations to parity with other providers

mariusvniekerk:
GitLab repos were effectively read-only: every mutation route returned `unsupported_capability` even though GitLab's API supports the operations.

- Post and edit comments on GitLab MRs and issues
- Open/close MRs and issues
- Merge MRs: `squash` maps to GitLab's squash flag, `merge` accepts under the project's configured merge method, and `rebase` returns a typed `unsupportedCapability` error (`merge_method_rebase`) since GitLab selects rebase behavior per project, not per merge
- Create issues and edit MR/issue title and body
- Approve MRs via the approvals API; a non-empty review body is posted as a regular note, and the synthesized approval event keeps an empty body so the text is not duplicated after sync
- Keep `request_changes` unsupported for GitLab with a typed `unsupportedCapability` envelope (`review_action_request_changes`)
- Preserve `thread_id` when event upserts edit threaded comments without an incoming thread id
- Surface GitLab merge HTTP errors as 409 conflicts when the provider reports an unmergeable MR
- Reject GitHub, Gitea, and Forgejo approval paths that cannot bind approval atomically to the reviewed head
- Require a current reviewed diff snapshot before head-bound merge/approve actions can pin a head SHA
- Preserve self-hosted GitLab host metadata on mutation error envelopes
- Keep PTY owner sessions alive until attachments that were active at natural process exit have drained
- Run tmux environment handoff scripts through `/bin/sh` so non-bash login shells do not break runtime launches
- Document GitLab's non-parity mutation semantics in `context/provider-architecture.md`


mariusvniekerk:
Addressed in 7bbefedd:

- **High (merge SHA binding):** `MergeMutator` now carries an `expectedHeadSHA`; the server passes the locally synced head (`PlatformHeadSHA`) and the GitLab client sets `AcceptMergeRequestOptions.SHA`, so GitLab rejects the merge with 409 when the source branch moved past the reviewed commit. The 409 surfaces as a typed `conflict` envelope and triggers a background MR resync.
- **High (approval SHA binding):** `ReviewMutator` carries the same `expectedHeadSHA`, set as `ApproveMergeRequestOptions.SHA`. The optional approval note is guarded by a head pre-check first so a stale review does not leave an orphaned comment (the approval itself is hard-bound by GitLab; mismatches map to a new `stale_state` platform code → 409 `conflict` + resync).
- **Medium (approval body invisible until next sync):** when a body was posted but the returned event carries none, the approve handler now syncs the MR inline, so the comment is present in the immediately-following local detail load — covered by an e2e assertion.

GitHub/Forgejo/Gitea currently treat the expected head SHA as advisory; enforcing it there is tracked as a follow-up since GitHub's sha-bound merge is a behavior change deserving separate validation. Validated with provider unit tests, wire-level e2e (stale merge/approve conflicts, immediate approval-comment visibility), and a full GitLab CE container run exercising the sha-bound merge and approval live.

mariusvniekerk:
Addressed in c896f107:

- **High (first-party paths don't pin / UI doesn't disable / omission accepted):** the server now rejects merge and approve on `mutation_head_binding` providers when `expected_head_sha` is omitted (`validationError` on `body.expected_head_sha`), so unbound mutations can no longer reach the provider. The approve button, palette approve action, and merge modal all send the rendered `platform_head_sha`, and head-bound actions are disabled (with an explanatory note in the merge modal) when the provider binds heads but no synced head is available. Coverage: Playwright asserts both first-party request bodies carry the rendered head; Go e2e asserts omitted pins are rejected for merge and approve; the GitLab CE container test now pins like a real client.
- **Medium (GitLab offered Rebase and merge):** GitLab project normalization now populates `MergeSettings` — squash per the project's `squash_option`, merge commit allowed, rebase never — so the modal no longer renders the rebase method GitLab cannot honor per merge.

mariusvniekerk:
Addressed the f909bb0 review's Medium in 7cf09aaa: the command-palette approve input now supplies onHeadConflict (reloading the detail with a sync-enabled load on stale_state/head_unknown, mirroring the detail-view buttons) and reads the rendered head from the detail envelope's canonical top-level platform_head_sha rather than the optional nested field.

mariusvniekerk:
On the 24a17a9 review: the Medium (head-conflict state surviving provider/host route changes) is fixed in the follow-up commit — the reset effect now tracks the full provider-aware identity. The High is the documented cross-provider enforcement gap: expected_head_sha is provider-enforced on mutation_head_binding providers (GitLab) and a local-cache guard elsewhere, as stated in context/provider-architecture.md and the capabilities matrix; threading the pin into GitHub's PullRequestOptions.SHA and Gitea/Forgejo's head_commit_id is the dedicated follow-up task already filed, since changing GitHub merge behavior deserves its own validation cycle rather than riding this PR.

mariusvniekerk:
Addressed the 000b526 review in a9be2139:

- **High (pin dropped before GitHub/Gitea/Forgejo provider calls):** merges now carry `expected_head_sha` to every provider API that gates on it — GitHub's `sha`, Gitea/Forgejo's `head_commit_id` (GitLab already sent `sha`) — with moved-head rejections classified as `stale_state`. Approvals are pre-checked against the live provider head on GitHub/Gitea/Forgejo before submission (their review commit ids record rather than gate; GitLab approval remains hard-gated), and GitHub reviews carry the pinned commit id. Covered by new provider tests on both paths.
- **High (live reactive pin while the form is open):** the merge modal and approve popover now capture the pin once at form-open; a background refresh cannot rebind an open form, and a genuinely moved head is rejected against the captured pin into the existing conflict recovery.

`context/provider-architecture.md` now states the exact per-provider strength of the pin.

mariusvniekerk:
Addressed the a9be213 review findings in 12be81f6:

- **High (approval TOCTOU)**: approvals on GitHub/Gitea/Forgejo now verify the live head again after the review submits; if it moved past the pin the approval is revoked upstream (GitHub review dismissal, Gitea/Forgejo review deletion) and the caller gets the 409 `stale_state` flow. A failed revocation is reported in the error rather than hidden. GitLab remains atomically gated by the approvals API `sha`.
- **Medium (ApproveButton reset)**: the popover reset effect now tracks `provider`, `platformHost`, and `repoPath` in addition to owner/name/number, and clears the captured pin; covered by a component test that routes the same numbers across providers and asserts the submitted pin.
- **Medium (gitealike 409 over-classification)**: only Gitea/Forgejo's "head target does not match" rejection classifies as `stale_state`; other 409s stay generic conflicts (unit-tested both ways).

Also added full-stack e2e coverage (HTTP + SQLite + recording GitHub mock) proving the reviewed head reaches the provider as merge `sha`/review `commit_id`, the moved-head rejection maps to 409 `stale_state`, and a moved-head approval is dismissed.

mariusvniekerk:
Addressed the 12be81f review (and the follow-up local review rounds) in 0303cee8 → 3d49c397 → bd5c295f:

- **High (hidden stale-approval side effects)**: 409 `stale_state` responses now carry the provider side-effect context — a failed revocation ("approval may stand…") or a GitLab note posted before the approval failed — in `details.context` and the human-readable `detail`, plus stable `details.revocation` (`succeeded`/`failed`) and `details.review_id` so clients branch without parsing prose. The contract (post-submit only; pre-check rejections omit the members) is pinned in `context/error-handling.md`.
- **Found while validating against a real Forgejo container**: both Gitea-like SDKs report any non-2xx merge response as `merged=false` with a *nil error*, so a rejected merge (stale head, conflict, permissions) was recorded locally as merged. A transport-level rejection capture now surfaces the real status/message, with the clear/request/read sequence serialized against concurrent merges. The live probe also showed current Gitea/Forgejo reject with "head out of date" (not the older "head target does not match"); classification matches both, and the container e2e now probes the live rejection shape end to end (validated green against Forgejo v12).
- Full-stack e2e added for: GitHub pin pass-through/stale mapping/dismissal, gitealike pinned-merge stale vs generic conflict, moved-head approval deletion, and failed deletion reporting `revocation: "failed"`.

mariusvniekerk:
Addressed the 7739d76 review finding in 1e57c972:

- **Medium (review content survives dismissal)**: a dismissed stale approval's context now states the posted review text remains visible upstream and a retry would repeat it, and the approve review-draft publish path wraps the post-submit stale error in `DiffReviewPublishPartialError` with the local comment ids — the server's existing partial-publish handling then clears the published draft copies and marks the response `partialPublish`. Content-less approvals keep the plain stale flow.
- Also added the concurrency regression coverage for the merge rejection capture (concurrent Gitea/Forgejo merges with distinct 409s, run under the race detector, no cross-talk).

mariusvniekerk:
Addressed the 1e57c97 review High in 5fb13a32 (with f01299de from the parallel session):

- **High (UI drops stale-approval context)**: the head-conflict callbacks now carry the problem's `details.context` from `runApprovePR` and the merge modal through `ApproveButton` into `PullDetail`'s persistent stale banner, so "approval may still stand" / "posted review text will be repeated" reaches the user instead of the generic re-review prompt alone. Covered by a keyboard-actions unit test for the forwarding and a Playwright spec asserting a failed-dismissal context renders next to the banner.
- f01299de additionally added the full-stack HTTP+SQLite coverage for the GitHub/GitLab review-draft partial stale-publish path (published draft comments cleared, 409 stale_state partial shape asserted).

mariusvniekerk:
Addressed the 960937a review (fail-closed verification landed in a69a1a3a from the parallel session; refinements in 3657978e):

- **High (fail open on unverifiable heads)**: post-submit approval verification on GitHub, the Gitea-likes, and the Forgejo draft path now fails closed — a head that cannot be read is treated like a moved head and the approval is revoked; both context docs state the rule.
- **Medium (Forgejo draft partial contract)**: failed review deletion on the Forgejo draft path wraps in `DiffReviewPublishPartialError` with the local draft comment ids and carries `revocation`/`review_id`.
- Follow-on design points from the same round: `details.cause` (`moved_head` | `head_unverifiable`) distinguishes the two stale flavors, and the partial-publish vs failed-revocation precedence is pinned — the 409 `stale_state` envelope wins, keeps the revocation members alongside `partialPublish`, and the server clears published local draft copies before responding. Proven through HTTP+SQLite e2e on both the failed-dismissal draft publish and the unverifiable-head approval.

mariusvniekerk:
Addressed the d5457b7 review findings in 67f8b56a:

- **High (non-GitHub sync never records a reviewed head)**: both the per-MR sync and the periodic list sync now compute and persist the diff snapshot (clone fetch + merge-base) for non-GitHub providers, mirroring the GitHub paths. Covered by a full-stack e2e that runs a plain provider sync against a real local git remote, asserts `reviewed_head_sha` is exposed, and merges with the sync-derived pin (`TestAPIGitealikeNormalSyncEnablesHeadBoundMutations`).
- **High (PATH-resolved rm in the workspace CWD)**: the pane cleanup now embeds an absolute `rm` resolved from the server's trusted PATH at script-generation time (`/bin/rm` fallback), asserted in the launcher test.
- **Medium (tmux flattens `/bin/sh -c` argv)**: the pane handoff now lives in an executable script file launched by a single dialect-neutral `exec /bin/sh '<script>'` token, so the bootstrap parses identically under POSIX shells and fish and never re-enters the user's default shell.

One pre-existing local-only failure remains tracked separately (macOS tmux exit-frame promptness; deterministic repro shows a bash link in the pane exec chain defeats tmux's fd-close EOF — CI skips tmux tests).

mariusvniekerk:
Addressed the 67f8b56 review findings in 107fa3f4:

- **High (GitLab still missing the reviewed head)**: `NormalizeDetailedMergeRequest` now fills base/head SHAs from `diff_refs` (the detail payload — the list endpoint omits it), so the reviewed diff snapshot writes on the detail sync. Covered by a GitLab-specific full-stack e2e (`TestGitLabNormalSyncEnablesHeadBoundMutations`) that runs a plain sync against a real local clone, asserts `reviewed_head_sha`, and merges with the sync-derived pin.
- **Medium (redundant clone fetches)**: list sync no longer re-runs `EnsureClone` per MR (the cycle already fetched the clone once); the snapshot is skipped entirely when the stored `DiffHeadSHA`/`DiffBaseSHA`/`MergeBase` already cover the current head/base pair, bounding periodic git work on large repos.
- **Medium (relative rm)**: the embedded cleanup `rm` is now required absolute (`filepath.IsAbs` guard on `LookPath`, `/bin/rm` fallback), so a relative `PATH` entry can never resolve a repo-controlled `rm` in the workspace CWD; regression test sets `PATH=.` with a hostile `./rm`.

mariusvniekerk:
Addressed the High from the 107fa3f review in 61cb3903:

- **High (list sync wiped the populated base SHA)**: `indexUpsertMergeRequest` now preserves a stored `platform_base_sha` when list data omits it and the head is unchanged (GitLab list payloads carry no `diff_refs`), matching the existing preserve-on-omit pattern for review decision / mergeable state / CI. Without it, the first periodic list sync after a detail sync blanked the base SHA, invalidated the diff snapshot, and re-armed 409 `head_unknown`. The GitLab sync-pin e2e now runs a second list sync after the detail sync and asserts both the snapshot and the merge pin survive.

The Low (`/bin/rm` fallback on NixOS) is already mitigated by the committed `filepath.IsAbs` guard and is being fully eliminated by the in-progress pane-bootstrap rework that removes the shell handoff entirely.

PR #482, opened 2026-06-11
Add middleman-github-app: GitHub App tokens for sync to relieve PAT rate limits

mariusvniekerk:
GitHub sync against github.com keeps exhausting the personal access token's 5,000 req/hour budget (shared with gh and other tooling), stalling every repo sync until the window resets. GitHub App installation tokens carry their own budget that scales with repository count, so a dedicated app takes middleman's high-volume polling off the PAT entirely — while writes stay on the PAT so merges and comments remain attributed to the user instead of an app bot.

User-visible changes:
- New `middleman-github-app` binary: `create` drives GitHub's App Manifest flow from the browser (one click to create, one to install) and writes the app id, private key, and installation into middleman's config as a `[[github_apps]]` entry; `list` shows installations and live rate budget; `install`/`uninstall`/`delete`/`open` manage the app afterward.
- middleman sync reads (REST, GraphQL bulk fetch, git fetch) authenticate with minted installation tokens ahead of PAT env vars; mutations keep using the user's own credential chain. The app manifest is read-only, since no app-authenticated code path writes.
- The browser steps render through an embedded Svelte setup page (create confirmation with permission summary, success view, stale-link state) in middleman's design language with dark mode.

Notable for review: the write path is a separate go-github client because go-github caches rate limits per client instance — a shared client would let an exhausted PAT (reported by a write response) preemptively block app-token reads, recreating the original failure. Mint failures surface as errors instead of silently falling back to the exhausted PAT. App deletion has no API, so `delete` verifies credentials work, has the user delete in browser settings, and polls for the credentials to die before removing local state.


mariusvniekerk:
All four review findings addressed in a8e26f6d:

- **viewer_can_merge from app responses (High)**: `GetRepository` now resolves with the write (user) credential — its `permissions` block is viewer-specific, and reading it with the read-only app token stored `viewer_can_merge=false` for users whose PAT can merge. The split-auth e2e now keys the fake's permissions on the credential and asserts `viewer_can_merge=true` lands in SQLite.
- **Write availability gated on the sync tracker (Medium)**: each GitHub host gets a `rest_write` tracker fed by mutation-path responses (and the settings refresh, which keeps it current); operation availability consults it when present, so app exhaustion no longer disables PAT writes and PAT exhaustion surfaces again. Covered by `TestAPIRepoResponseOperationsGateOnWriteTrackerWhenSplit`.
- **repository_selection ignored (Medium)**: selected installs are verified against `GET /installation/repositories` before being recorded; missing repos (and glob patterns, which can't be enumerated) reject the install with guidance. Covered by accept/reject e2e tests.
- **installation_id without installation_account (Medium)**: now a config validation error instead of silently skipping coverage validation.

mariusvniekerk:
Second review round addressed in 0832cffb:

- **Write trackers on every host (High)**: trackers and the credential split are now wired only for hosts with an active `[[github_apps]]` installation. PAT-only hosts keep gating writes on the sync trackers, and their write responses feed the shared tracker exactly as before the split existed.
- **GraphQL write bucket (Medium)**: ready-for-review was cataloged as REST but is GitHub's GraphQL-only mutation; it now has a `graphql_write` tracker fed from its response headers and gates on the GraphQL write bucket independently of REST exhaustion.
- **GetRepository fully on PAT (Medium)**: metadata reads are back on the sync credential so app-token-only hosts keep syncing; split hosts overlay the viewer-specific permissions from the PAT, and a failed overlay degrades `viewer_can_merge` to unknown rather than failing sync or storing the app's read-only permissions.
- **Path safety (Medium)**: the slug from the manifest conversion response is validated against GitHub's slug shape before use as a filename (with a containment check), and the key path is stored absolute — which also fixes relative `--config` paths double-resolving into `tmp/tmp/`. Covered by a hostile-slug unit test and a nested-relative-config e2e.

Also from the per-commit reviews: `/flow.json` dies once the callback consumes it (no manifest replay from a refreshed tab), repo-level token overrides no longer fall through to the app token when unset, `list` now verifies selected-repository installs against configured repos for manually edited configs, and a Playwright suite drives the built Svelte setup page in a real browser (3 tests: hand-off, auto-continue, stale link).

cpcloud:
Guessing users still need a PAT with write permissions on various things to execute actions (merging PRs, commenting, etc) that mutate the repo in some way?

That seems fine, given that the volume of those actions is typically human-scale, though that's probably not a long-term reliable fact.

mariusvniekerk:
The user INTENTIONALLY still needs a PAT for write permissions since otherwise those actions will show up in the UI as the bot,

cpcloud:
Possible to bring-your-own-app to this? Just wondering, no need to clank on it here if it doesn't work out of the box.

mariusvniekerk:
probably, since you just need the config for things to work,  the helper app is just a way to make one.

mariusvniekerk:
All substantive review findings through the latest combined reviews are addressed (branch rebased onto main twice along the way, most recently picking up the Host-validation middleware and the assignee/reviewer feature):

- **Issue sync response** carries `repo.operations` again, so a sync cannot clear the frontend's mutation gates (`d3e14eac`).
- **`install` is now a working repair path**: the CLI loads configs with only the app-coverage validation relaxed (`LoadForGitHubAppRepair`), refreshes a recorded installation's `repository_selection`/`selected_repos` in place, and falls through to waiting for an installation on the owning account when the recorded one is on the wrong account or gone. Stale-snapshot and wrong-account repairs both have e2e coverage (`d3e14eac`, `718c702a`).
- **`create_issue` operation** added; the repo summary's New issue button disables with the server's reason like every other mutation (`d3e14eac`).
- **Assignee/reviewer mutations** ride the write credential (PAT) with split-auth e2e coverage of the add and remove paths (`745db3f3`, `9e47dd44`).
- **Private keys** are written with `O_EXCL`, so an existing file or symlink is never silently replaced (`d3e14eac`).
- **Loaded-draft availability flips** are covered by a mounted split-view round trip driven by the real detail poll rather than a page reload (`718c702a`).

Two recurring security findings are intentionally not being addressed, as they assume an attacker with local code execution on the user's own machine: the `/flow.json` loopback "hijack" during the seconds-long setup window, and the "JWT leak" via the user passing `--api-base` to their own CLI. Both require the attacker to already be in a position where far simpler attacks (reading the key file the CLI just wrote with the same UID) are available.

mariusvniekerk:
> Possible to bring-your-own-app to this?

Yes — nothing assumes the app was created by the CLI. Two ways:

- Hand-write the `[[github_apps]]` entry: `host`, `app_id`, `private_key_path`, `installation_id`, `installation_account`. Token minting only needs the app ID + private key + installation ID.
- Or write a minimal entry (`host`, `app_id`, `private_key_path`) and run `middleman-github-app install` — it signs with your existing key, picks up the installation, verifies repo coverage, and records the installation details and repository selection for you. `list` works against any configured app too.

The only requirement is that the app grants the read permissions sync touches: `contents`, `issues`, `pull_requests`, `checks`, `statuses`, `actions`, `metadata` (all read — writes always use your own credential). Missing permissions aren't validated up front; they'd surface as 403s on the affected sync endpoints.

PR #487, opened 2026-06-11
Fix Kata daemon switcher behavior

cpcloud:
- Keeps the Kata daemon selector visible in the header even when the roster has one daemon.
- Keeps daemon connection and health details inside dropdown rows instead of showing stale header text like `Connecting - kenn`.
- Treats local Kata daemons as tokenless loopback targets across catalog loading, health checks, and proxying.
- Shows real Kata request failures as a visible workspace alert instead of hiding them in the daemon menu.


cpcloud:
Screenshot:

image

cpcloud:
Not my favorite ux, but workable, and should get rid of the local "authentication required" issue.

mariusvniekerk (approved):

mariusvniekerk (approved):

PR #493, opened 2026-06-12
fix: stop background kata refreshes from swallowing row clicks

mariusvniekerk:
Clicking a task row in Kata mode while a background view refresh was in flight (a live SSE event, the bootstrap event-cursor sync, or a mutation reload) left the detail pane stuck on "Select a task". The refresh captured its preferred selection before its view fetch started, so on completion it reloaded the detail with that stale UID — aborting the click's in-flight detail fetch, marking it superseded, and clearing the pending selection. On daemons with busy agents there is almost always a refresh mid-flight, so clicks looked dead. (The list's keyboard-select debounce was a red herring; clicks bypass it.)

`refreshCurrentView` now treats user selection intent as newer than itself: an in-flight detail load keeps ownership of the pane, and a selection that completed during the view fetch is re-resolved instead of reverted. Explicit navigations (view/scope/route/daemon changes) still supersede pending selections.

- Clicking (or keyboard-selecting) a task now reliably opens it, even while live updates are streaming in

### DIFF
--- a/frontend/src/lib/features/kata/KataWorkspaceEventStream.test.ts
+++ b/frontend/src/lib/features/kata/KataWorkspaceEventStream.test.ts
@@ -101,6 +101,89 @@ describe("KataWorkspace", () => {
     });
   });
 
+  it("keeps a row selection made while a stream-triggered reload is in flight", async () => {
+    let streamController: ReadableStreamDefaultController<Uint8Array> | undefined;
+    vi.spyOn(globalThis, "fetch").mockImplementation(async (input) => {
+      const url = new URL(input instanceof Request ? input.url : String(input), window.location.origin);
+      if (url.pathname === "/api/v1/kata/daemons") {
+        return Response.json({
+          daemons: [
+            {
+              id: "home",
+              url: "http://127.0.0.1:7777",
+              default: true,
+              auth: "none",
+              health: "connected",
+            },
+          ],
+        });
+      }
+      if (url.pathname === "/api/v1/kata/proxy/api/v1/events/stream") {
+        return new Response(
+          new ReadableStream<Uint8Array>({
+            start(controller) {
+              streamController = controller;
+            },
+            cancel() {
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
Validated both review findings against the code:

**`kata-workspace.svelte.ts:808` (stale detail reload) — confirmed real, fixed in 19bdc6ff.** The unconditional `detailAbort` guard did preserve in-flight loads issued by an *older* refresh, and because the early return never bumped `detailRequestID`, the stale pre-mutation payload would later apply and leave both the pane and the cached ETag at the pre-mutation revision (so the next edit would send a stale `If-Match`). The exact scenario is reproduced as a store test (`a mutation refresh supersedes a stale detail reload from an older event refresh`), which fails against e584a3eb and passes with the suggested epoch gate (`detailRequestID !== selectionEpoch`). Both original click-swallow regressions still pass with the gate in place.

**`KataWorkspaceEventStream.test.ts:104` (e2e coverage) — not actionable at the browser-e2e layer.** The race requires deterministically holding a daemon view fetch open between the SSE event and the row click. The full-stack e2e harness (real middleman server proxying a fixture daemon catalog via `KATA_HOME`) has no hook to stall an upstream daemon response, so a browser e2e of this race would either pass vacuously when the timing doesn't line up or need a purpose-built stallable daemon fixture. The workspace test drives the real SSE wire path (actual `text/event-stream` bytes through `readKataEventStream`), the real store, and the real component tree — only the HTTP API boundary is mocked, which is the same boundary an e2e harness would have to control to make the race deterministic. The non-race behavior (row click → detail load, live stream reloads) is covered by the existing full-stack suites: kata.spec.ts (142 passed), kata-live.spec.ts + navigation.spec.ts (14 passed, 8 live-daemon-gated skips), all run locally against this branch.

Validation for 19bdc6ff: full frontend unit suite (2273 passed) plus the three Playwright suites above.

PR #495, opened 2026-06-12
fix: dismiss assignee/reviewer pickers on outside click

mariusvniekerk:
The assignee and reviewer chip popovers only closed via their X button or re-clicking their own chip, so a click elsewhere on the page left the panel floating, and pressing one chip while the other picker was open stacked both panels on screen.

UserListEditor now uses the same document-mousedown dismissal pattern the label picker already follows: any press outside the chip and panel closes the picker. That one mechanism also makes the pickers mutually exclusive, since pressing another editor's chip counts as an outside press for the open one.

- Clicking outside an open assignees/reviewers picker now dismisses it
- The assignees and reviewers pickers can no longer be open at the same time (also mutually exclusive with the label picker)

Demo:

https://github.com/user-attachments/assets/67bab759-3087-4712-9e2d-9c126e662bd8

### DIFF
--- a/frontend/src/lib/components/detail/UserListEditor.test.ts
+++ b/frontend/src/lib/components/detail/UserListEditor.test.ts
@@ -100,6 +100,28 @@ describe("UserListEditor", () => {
     expect(screen.queryByRole("dialog", { name: "Edit assignees" })).toBeNull();
   });
 
+  it("closes an open picker when another editor's chip is activated by keyboard", async () => {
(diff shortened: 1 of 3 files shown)


mariusvniekerk:
Demo of the fix (outside-click dismissal, then picker swap):

https://github.com/user-attachments/assets/67bab759-3087-4712-9e2d-9c126e662bd8

PR #497, opened 2026-06-12
Fix added diff highlighting regressions

mariusvniekerk:
Added-file diffs were still exposed to two regressions after the initial highlighting fix: hunk-only fallback patches could lose meaningful trailing whitespace, and complete added Go files with long signatures did not have browser-level coverage for the syntax-enabled render path.

This preserves raw hunk-only patch whitespace, adds the missing browser regression for complete added Go files, and keeps the frontend check hook reliable under local memory pressure by splitting the check phases with explicit heap caps.

Validation:
- `./node_modules/.bin/vp test packages/ui/src/components/diff/pierre-diff.test.ts packages/ui/src/components/diff/pierre-worker-pool.test.ts`
- `make frontend-check`
- `cd frontend && bun x playwright test --config=playwright-e2e.config.ts --project=chromium tests/e2e-full/diff-view.spec.ts`


mariusvniekerk:
> packages/ui/src/components/diff/pierre-diff.ts:143 

TRULY TRULY insane, from the security reviewer.

PR #498, opened 2026-06-12
Remove Mermaid viewer pan controls

cpcloud:
- Removes the directional Mermaid diagram pan buttons from inline and expanded viewers.
- Keeps the reset button available for restoring the diagram viewport after drag panning or wheel zooming.
- Updates browser coverage so the viewer exposes reset without pan controls.


cpcloud:
Merging on green.

PR #500, opened 2026-06-13
test(frontend): move non-browser e2e coverage from Playwright to Vitest

mariusvniekerk:
The mock-API Playwright lane paid full-browser cost — two browser projects plus a Vite dev server per run — for behavior that needs no real rendering: keyboard dispatch, focus, store-driven DOM, fetch request shapes, and URL routing. That coverage now runs in the Vitest jsdom lane that already executes on every change, mounting the real app shell (real Provider, stores, views, keyboard dispatch) with the API mocked at the fetch boundary. A roborev finding on the first commit (the jsdom mock began as a copy of the Playwright mock) is addressed in the second: both suites now consume one fixture/route core, so they cannot drift.

## Changes

- Converted 9 mock-suite specs (36 cases) into 37 Vitest cases: activity collapse, activity thread runs, budget display, issue routing, palette commands, palette focus trap, PR-detail palette commands, cheatsheet, keyboard shortcuts.
- Folded the two `tests/e2e-full` duplicates (activity collapse, budget display) into the same Vitest files and deleted them.
- Added `src/test/appHarness.ts` (mounts the real `App.svelte` in jsdom) and made `src/test/mockApiFetch.ts` the single mock-API source; `tests/e2e/support/mockApi.ts` is now a thin `page.route` adapter over the shared handler.
- Specs that genuinely need a browser (screenshots, video, geometry, computed CSS, mouse drag) stay in Playwright.

## Runtime

Measured back-to-back on the same machine (matched load):

| Lane | What runs | Wall time |
|---|---|---|
| Before: Playwright (9 specs) | 36 cases × chromium + firefox, incl. Vite dev-server boot | 22s |
| After: Vitest (same coverage, 9 files) | 37 cases | ~4.5s |

The converted cases also join the `vp test` lane that already runs on every change, so their marginal cost there is near zero (full suite stays ~12s wall for 2,318 tests), and CI sheds the browser provisioning and dev-server boot for this coverage entirely.

Validation: full Vitest suite 2,318 passed; full mock Playwright suite after the final edit 320 passed / 8 pre-existing skips; typecheck identical to clean-HEAD baseline.


mariusvniekerk:
Okay this one is basically good to go now.

PR #501, opened 2026-06-13
Federated fleet: cross-host snapshot, HTTP/SSH peers, thin-client daemon contract

wesm:
## Summary

Adds a fleet federation layer so one middleman daemon (the **hub**) merges its own state with the state of remote middleman daemons (**peers**) into a single snapshot, proxies mutations to the daemon that owns the resource, and hands clients native terminal attach commands for sessions anywhere in the fleet. Every daemon runs the same binary; hub vs. peer is purely which daemon a client talks to and which peers its config lists.

Also removes the in-process embed library API. Middleman is now a standalone daemon reached by thin HTTP clients, not an embeddable Go library.

`docs/federated-fleet.md` is the single reference for the subsystem.

## What this adds

**Read plane.** `GET /api/v1/snapshot?include_peers=true` fans out to every configured peer's `/api/v1/snapshot/raw` concurrently (bounded per peer by `fleet.peer_timeout`) and merges into one enriched snapshot. A slow or unreachable peer degrades to `reachable: false` with a diagnostic instead of failing the merge. `/api/v1/snapshot/raw` is this host only and never fans out.

**Write plane.** Hub-keyed routes under `/api/v1/fleet/hosts/{host_key}/...` resolve to the local handler, an HTTP reverse proxy, or the SSH relay. Covers workspace and runtime-session create/delete, registered-worktree runtime, project register/delete, and filesystem probes.

**Terminal plane.** `.../attach-spec` returns the native command vector for a tmux-backed session; for SSH peers the command is wrapped in the peer's ControlMaster invocation so it runs from the hub host. WebSocket terminal routes proxy interactive sessions for local and HTTP-peer hosts and return an explicit 501 (pointing at the attach-spec) for SSH peers.

**Live tmux monitoring.** Each daemon samples its own tmux server (4s inventory, 15s metrics, bounded timeouts) and reconciles live sessions against DB ownership rows; unmanaged sessions are redacted to summary unless `[fleet.tmux] include_unmanaged_details` is set.

**Transports.** Peers are reached over HTTP (`[[fleet.peers]]`) or SSH (`[[fleet.ssh_peers]]`). For SSH the hub owns one ControlMaster per peer (`internal/sshfleet`) and relays each API call by executing the peer's `middleman api` verb remotely, so the peer's HTTP listener never leaves its host. A relay that finds no daemon on the peer starts one (`nohup middleman serve`) and retries once.

**Daemon contract for thin clients.** Runtime discovery via the BSD flock on `<data_dir>/middleman.lock` plus `middleman.run.json` (listen address, base path, token path); a bearer token minted to `<data_dir>/auth_token` (0600), gated on by `[api] require_auth`; the `middleman api [-i] [-d @-] METHOD PATH` CLI verb that discovers the daemon, authenticates, and relays one request with an exit-code contract (0/1/2). `PUT /api/v1/ui/active-worktree` records client UI focus; `GET`/`PUT /api/v1/settings/fleet/ssh-peers` lets clients read and replace the SSH peer set (validated, persisted, `restart_required` reported).

## What this removes

The root `middleman` package (`New`/`Options`/`Instance`, `EmbedHooks`, `EmbedConfig`, and the `FleetRemoteSource`/`FleetWriteForwarder`/`FleetIdentity` server hooks) is deleted: middleman is a standalone daemon, and the in-process library/embedding surface is unused. `ServerOptions` loses those fields; the served bootstrap config narrows to the daemon-owned `ui.activeWorktreeKey`; the host-check `Disabled` escape hatch is gone.

## Code organization

`fleet/` moves to `internal/fleet/` — it was public only for the removed library surface, and the repo otherwise keeps everything under `internal/`.

## Database

One new migration, `000033_fleet_integration`: registered-worktree and host-scoped tmux session tables, a worktree-stats table, and project/worktree discovery columns (`is_stale`, `repository_kind`, `is_hidden`, `session_backend`, `linked_issue_numbers`). Per repo policy (no migration compatibility before first release) this is a single delta against `main`.

## Config surface

```toml
[fleet]
key = "studio"          # this daemon's fleet identity
peer_timeout = "2s"     # per-peer snapshot fetch budget
[fleet.tmux]
include_unmanaged_details = false
[[fleet.peers]]         # HTTP peer
key = "mini"
base_url = "http://mini.tail:8091"
[[fleet.ssh_peers]]     # SSH peer (no exposed listener)
key = "epyc"
destination = "user@epyc.tail"
[api]
require_auth = true
```

## Testing

Wire-level tests per fleet route (degraded-peer fan-out, cold-SSH-peer single-flight, attach-spec wrapping, daemon auto-start), `internal/sshfleet` unit tests for the ControlMaster lifecycle and relay framing, `cmd/middleman` e2e against the real binary (auth, framing, exit codes, base-path), `internal/server/apitest` coverage through the generated client for the settings routes, and a Docker-backed container e2e (`fleet_container_e2e_test.go`) running a real hub + member.

## Notes for reviewers

- Of ~57k inserted lines, ~29k are regenerated OpenAPI/client artifacts (`internal/apiclient/generated`, `frontend/openapi/openapi.yaml`, `packages/ui/src/api/generated/schema.ts`); hand-written change is ~28k. Generated artifacts come from `make api-generate`.
- A daemon already running on this branch's pre-consolidation schema needs a one-time `schema_migrations` version reset (or a fresh data dir), since the migrations were consolidated to the single `000033` delta.


wesm:
will rebase

mariusvniekerk:
The High findings is pure security theater.

mariusvniekerk:
remaining findings are irrelevant.

PR #503, opened 2026-06-14
fix: launch tmux panes without a shell handoff (fish/Nix-safe)

mariusvniekerk:
main's tmux pane handoff builds a multi-line POSIX shell snippet (source an env file, `/bin/rm` cleanup, then `exec env -i KEY="${KEY-}" ... cmd`) and runs it through the user's default shell. That breaks two real setups:

- **fish users**: the POSIX snippet doesn't parse under fish, so pane launch fails.
- **NixOS users**: the hardcoded `/bin/rm` cleanup fails (no `/bin/rm`).

This replaces the shell snippet with a hidden `__tmux-pane-bootstrap` subcommand: the launcher writes the pane env and argv to a 0600 handoff file and emits a single dialect-neutral token, `exec '<exe>' __tmux-pane-bootstrap '<handoff>'`. The new `internal/workspace/panebootstrap` package re-execs the middleman binary, applies the handoff env (env -i semantics), and execs the command directly.

- Pane launch works under fish and on NixOS (no shell snippet, no `/bin/rm`)
- No shell in the pane exec chain; preserved env values still kept out of tmux argv
- Forecloses the macOS tmux pane-teardown regression a `/bin/sh` handoff reintroduces (PTY EOF, not process exit, must drive the terminal's exit frame)
- Reuses the `os.Executable()` self-re-exec idiom already used by `internal/ptyowner` for its pty-owner subcommand

### DIFF
--- a/cmd/middleman/main.go
+++ b/cmd/middleman/main.go
@@ -34,6 +34,7 @@ import (
 	"go.kenn.io/middleman/internal/telemetry"
 	"go.kenn.io/middleman/internal/tokenauth"
 	"go.kenn.io/middleman/internal/web"
+	"go.kenn.io/middleman/internal/workspace/panebootstrap"
 )
 
 type splitLogHandler struct {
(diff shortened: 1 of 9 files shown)


mariusvniekerk:
abandoning.

PR #507, opened 2026-06-15
fix: restore provider approval reviews

mariusvniekerk:
The GitLab parity work made approval availability depend on providers proving an atomic head-bound approval path. That hid the APPROVE button for GitHub and under-advertised Gitea-like provider review APIs, even though approval is a safe review mutation when middleman sends a head SHA opportunistically instead of requiring one locally.

This restores approval as a normal provider review action while keeping merge on the stricter head-bound path. Reviewers should treat this as a regression fix for maintainer workflow parity across providers, with GitLab approval matching its API contract where `sha` is optional but checked when supplied.

Validation:
- `make test-short`
- `make frontend-check`
- `cd frontend && bun x playwright test --config=playwright.config.ts --project=chromium tests/e2e/head-pinned-actions.spec.ts`
- `cd frontend && bun x playwright test --config=playwright.config.ts --project=firefox tests/e2e/head-pinned-actions.spec.ts`
- `cd frontend && bun x playwright test --config=playwright-e2e.config.ts --project=chromium tests/e2e-full/detail-action-buttons.spec.ts`
- `cd frontend && bun x playwright test --config=playwright-e2e.config.ts --project=firefox tests/e2e-full/detail-action-buttons.spec.ts`
- focused provider/server approval tests
- frontend approval action tests

### DIFF
--- a/frontend/src/App.svelte
+++ b/frontend/src/App.svelte
@@ -784,9 +784,10 @@
     const workflowApprovalReady = Boolean(
       capabilities.workflow_approval && wfa?.checked && wfa.required,
     );
-    // Same rendered-head source as PullDetail: the detail envelope's
-    // verified reviewed_head_sha, not the mutable platform_head_sha.
+    // Merge remains tied to the verified reviewed_head_sha, but approval
+    // should include the latest synced provider head when one is known.
     const renderedHeadSha = detail.reviewed_head_sha ?? "";
+    const latestPlatformHeadSha = detail.platform_head_sha ?? "";
     // TS cannot carry the !stores narrowing into the closure below.
     const appStores = stores;
     return {
@@ -794,7 +795,7 @@
         State: pr.State,
         IsDraft: pr.IsDraft,
         MergeableState: pr.MergeableState,
-        platform_head_sha: renderedHeadSha,
+        platform_head_sha: latestPlatformHeadSha,
       },
       ref: {
         provider: sel.provider,
--- a/frontend/tests/e2e-full/detail-action-buttons.spec.ts
+++ b/frontend/tests/e2e-full/detail-action-buttons.spec.ts
@@ -414,12 +414,12 @@ test.describe("detail action buttons", () => {
     const merge = page.locator(".btn--merge");
     const close = page.locator(".btn--close");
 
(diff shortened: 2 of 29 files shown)


mariusvniekerk:
@roborev-ci The reviewed safety gate is WRONG and breaks functionality

mariusvniekerk:
On the approval head-binding concern: the UI approval flow intentionally binds to the current synced provider head, not `reviewed_head_sha`.

For the direct Approve button there is no drift between opening the dialog and submitting it: `ApproveButton` snapshots the loaded `platform_head_sha` into `pinAtOpen` when the dialog is opened, and the submit path sends that frozen value as `expected_head_sha`. A background detail refresh while the dialog is open cannot silently rebind the approval target.

For palette approval there is no intermediate dialog, so it sends the currently loaded `platform_head_sha` at command execution. Merge remains stricter and is still gated on `reviewed_head_sha`.

So the review finding is only valid under a different contract, namely "approval must target the reviewed diff head." That is not the intended contract here. The intended contract is "approve the current synced provider head at the moment approval is started," and the button/palette paths now do that.

mariusvniekerk:
Follow-up on the approval head-binding roborev comments:

The main finding assumes direct approval must target the verified reviewed diff head (`reviewed_head_sha`). That is not the intended contract for this branch. Direct `/approve` targets the provider head captured when approval starts: the toolbar freezes the loaded `platform_head_sha` when the dialog opens, and the palette sends the currently loaded `platform_head_sha` at command execution. Merge and review-draft publish remain stricter and continue to use the verified reviewed diff head.

The valid residue in the review was stale wording around that contract. `ff0febb` updates the endpoint, capability, provider, draft-publish, and e2e comments so `expected_head_sha` is described as endpoint-specific: provider/platform head for direct approval, reviewed diff head for merge/draft publish. It also documents omitted approval pins as an older-client compatibility fallback and stale supplied approval pins as delegated to provider head-binding where available.

The behavior is covered by divergent-head browser e2e and full-stack API/SQLite coverage: toolbar approval and palette approval send `platform_head_sha`, merge stays pinned to `reviewed_head_sha`, and omitted direct approval pins bind to the stored provider head.

PR #516, opened 2026-06-16
Fix Markdown rich diff rendering

mariusvniekerk:
## Summary
- Render Markdown rich preview as an annotated unified document instead of forcing before/after panes.
- Use the same rendered diff tree for split mode, with block placeholders so paragraphs stay aligned.
- Serialize frontend/UI Svelte checks through a single Vite+ Svelte-check task to avoid overlapping diagnostics.

## Screenshots
Unified:

Split:


mariusvniekerk:
we do NOT need an e2e test for this.  this is very much a component thing.

PR #521, opened 2026-06-16
Run frontend tooling through bun-latest

mariusvniekerk:
The frontend Vite+ paths still depended on Node-launched Vite+ or the global vp shim, which made local installs and process-compose startup sensitive to whatever runtime happened to be first on PATH. That defeated the goal of using the bun-latest build installed with bun-pr and also let test commands resolve dependencies from the wrong Vite+ installation.

This switches the package scripts, Vite+ task commands, build helpers, and process-compose frontend entrypoint to run the repo-local Vite+ binary through bun-latest. The process-compose path now goes through the shared frontend script so sparse environments recover the mise-managed bun-latest path, while still exposing node because Vite+ internals require a node binary to be discoverable. Installs use --no-save so dependency setup does not mutate package.json during startup.

Validation run:
- make frontend
- sh -n scripts/dev-stack-frontend.sh
- bash -n scripts/frontend-dev.sh
- git diff --check
- stripped-PATH sh ./scripts/dev-stack-frontend.sh launched Vite+ successfully on a test port
- bun-latest run --cwd frontend test passed 237 files / 2362 tests with 2 skipped before the final script-path changes
- node node_modules/.bin/playwright test --config=playwright.config.ts was run and failed in existing UI specs unrelated to the runtime/install paths: inline-review still showed Publish review after submit_review flipped unavailable, and workspace sorting matched Activity against both Activity and Item activity in strict mode

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,9 +24,9 @@ AIR_BIN := $(shell if command -v air >/dev/null 2>&1; then command -v air; \
 DEV_LOG_DIR ?= tmp/logs
 DEV_BACKEND_LOG ?= $(DEV_LOG_DIR)/backend-dev.log
 VITE_PLUS_VERSION := 0.1.24
-VITE_PLUS_BIN := node ./node_modules/vite-plus/bin/vp
-VITE_PLUS_FRONTEND_BIN := node ../node_modules/vite-plus/bin/vp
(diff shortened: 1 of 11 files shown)


mariusvniekerk:
This is not likely to ever merge, this was an experiment to see how things are under the rust based bun (unreleased)

PR #542, opened 2026-06-18
ci: run e2e in the Playwright container to end firefox install flakiness

mariusvniekerk:
The firefox e2e leg installed browsers with `playwright install --with-deps`, pulling ~40 OS packages off the apt mirror — which repeatedly crept against the 10m step cap and timed out (chromium, ~9 packages, never did). Reruns made it a coin flip.

This runs the e2e job inside `mcr.microsoft.com/playwright`, where browsers and their OS deps are pre-baked, so the download and the apt install disappear for every browser. The flaky step is removed, not retried.

- e2e job now runs in `container: mcr.microsoft.com/playwright:v1.60.0-noble`
- removed the per-leg `playwright install --with-deps` step and the now-redundant ms-playwright cache
- new pre-commit guard (`make playwright-version-check`) fails the commit if the container tag and the pinned `@playwright/test` version ever drift apart

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -288,6 +288,16 @@ jobs:
           cache: true
           run-install: true
 
+      - name: Install tmux for workspace PTY sessions
+        # The minimal Playwright image ships browsers but not tmux, which the
+        # full-stack workspace e2e specs need to start real PTY sessions
+        # (ubuntu-latest had it pre-installed). This is one small package, not
+        # the 48MB firefox browser-deps download the container already removed.
+        timeout-minutes: 5
+        run: |
+          apt-get update
+          apt-get install -y --no-install-recommends tmux
+
       - name: Build frontend
         run: cd frontend && vp build --logLevel warn
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,6 +261,12 @@ jobs:
 
   e2e:
     runs-on: ubuntu-latest
+    # Run inside the official Playwright image: browsers and their OS deps are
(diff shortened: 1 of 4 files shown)


mariusvniekerk:
Given that the tests PASS in CI, roborev is wrong here.

.github/workflows/ci.yml
```diff
@@ -261,6 +261,12 @@ jobs:
 
   e2e:
     runs-on: ubuntu-latest
+    # Run inside the official Playwright image: browsers and their OS deps are
+    # pre-baked at $PLAYWRIGHT_BROWSERS_PATH, so there is no apt download to
+    # time out on (the firefox `--with-deps` install repeatedly flirted with the
+    # old 10m cap off a slow mirror). The image tag is version-coupled to the
+    # @playwright/test pin and kept honest by `make playwright-version-check`.
+    container: mcr.microsoft.com/playwright:v1.60.0-noble
```
cpcloud: ideally a hash here. bots can take care of keeping it up to date.

mariusvniekerk: Done in 95c3a6ba — pinned to `mcr.microsoft.com/playwright@sha256:9bd26ad9...946b948` (the multi-arch index digest for `v1.60.0-noble`), matching the SHA-pinning style of the `actions/*` above. Kept a trailing `# v1.60.0-noble` comment so the version stays human-readable, and the `playwright-version-check` guard now reads the version from that comment: it fails the commit if the version drifts from the `@playwright/test` pin, and also if the comment is ever dropped — so a bot bumping the digest can't leave it untraceable.

cpcloud (approved):

PR #547, opened 2026-06-18
fix: keep filtered diff scrolling local

mariusvniekerk:
Passive scrolling in the PR diff should update the current-file highlight without treating that highlight update as a command to reveal and focus rows in the file tree. Under file-category filters like Code, that coupling made normal wheel scrolling look like the diff was reloading or remounting even though no new diff data was needed.

This keeps the diff pane scrollable, keeps the filtered tree and diff body aligned, and reserves file-tree reveal behavior for explicit navigation such as sidebar selection, keyboard file jumps, or line jumps.

Validation: full frontend unit suite, UI package check, Svelte analyzer on touched files, and live Chrome verification of Code-filter wheel scrolling with no extra diff/files requests, route changes, remounts, or file-tree reveal calls.

Visual capture was kept local because the available reproduction uses live middleman data and the screenshot upload policy requires explicit approval before attaching live captures.

### DIFF
--- a/frontend/src/lib/components/settings/RepoSettings.test.ts
+++ b/frontend/src/lib/components/settings/RepoSettings.test.ts
@@ -408,6 +408,7 @@ describe("RepoSettings", () => {
         renderer: "xterm",
       },
       agents: [],
+      fleet: defaultFleetSettings(),
     });
     mockUpdateRepoWorktreeBasePath.mockResolvedValue({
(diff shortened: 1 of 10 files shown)


mariusvniekerk:
An explicit jump targeting a hidden file SHOULD BE IMPOSSIBLE

cpcloud (approved):

PR #548, opened 2026-06-19
fix: clarify eager refresh budget display

cpcloud:
## Summary

- Clarifies the eager refresh budget as budgeted sync work for details, comments, and backfills, and stops counting GitHub 304 responses against that budget.
- Keeps the bottom status bar focused on REST and GraphQL pressure while leaving eager refresh budget detail in the API Budget popover.
- Tightens the over-budget popover copy so it explains the paused work without an extra deferred fragment.

### DIFF
--- a/frontend/src/lib/components/layout/BudgetBars.svelte
+++ b/frontend/src/lib/components/layout/BudgetBars.svelte
@@ -95,7 +95,7 @@
 
   {#if b.hasAny}
     <span class="budget-count" style:color={syncBudgetColor(b.spent, b.limit)}>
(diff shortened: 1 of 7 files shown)


mariusvniekerk:
Should we have this number be configurable?

cpcloud:
You mean the eager refresh budget?

cpcloud:
I'm sort of on the fence of whether it's even useful to show users, since it's not really actionable without a new config value, or some kind of CTA like "refresh now for the freshest bytes"

cpcloud:
This PR was mostly about surfacing it in a way that clarifies what it's doing/gating/measuring.

PR #553, opened 2026-06-19
Clarify frontend test lane selection in project guidance

mariusvniekerk:
## Summary
- Clarify that Playwright and full-stack e2e are not mandatory for every visible frontend change.
- Add explicit lane guidance for Vitest component tests, Vitest browser tests, Playwright, and backend/server tests.
- Note that a real backend test plus a component or browser test can be enough when they cover the runtime path and presentation separately.

## Testing
- Not run (not requested)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -119,10 +119,12 @@ make vet        # go vet
 
 ### End-to-End Tests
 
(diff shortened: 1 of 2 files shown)


cpcloud (approved):
The agents DO love their playwright. One line CSS change? Time to spin up an entire browser stack.

PR #559, opened 2026-06-19
Show user avatars in picker rows

mariusvniekerk:
- Show provider profile images in reviewer and assignee pickers when they load successfully.
- Keep initials as the fallback for manual users and accounts whose image URL fails.

Screenshot:


mariusvniekerk:

PR #567, opened 2026-06-22
Keep activity detail open after merge modal escape

wesm:
- Escape now dismisses the merge modal without also closing the Activity PR detail split.
- Adds e2e coverage for the Activity split + merge modal Escape flow.


wesm:
tested and working for me

PR #574, opened 2026-06-22
ci: split Go and frontend unit tests

mariusvniekerk:
The CI test lane was doing Go and frontend unit work in one job, which made the check heavier than necessary and blurred which stack actually failed. Splitting the lane lets Go tests run without the frontend toolchain setup, while frontend unit failures report through their own focused status.

This intentionally does not add an aggregate status job. Reviewers and branch protection should key off the real Go and frontend test jobs instead of a wrapper that performs no verification.

Workflow syntax was checked locally with actionlint and yq, and the repository commit hooks passed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -59,7 +59,8 @@ jobs:
       - name: Run NilAway
         run: make nilaway
 
-  test:
+  test_go:
+    name: Go tests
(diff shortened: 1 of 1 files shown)


mariusvniekerk:
Reviewer is overly cautios since there are no branch protections needing that.

PR #575, opened 2026-06-22
Clarify workspace onboarding and launch targets

mariusvniekerk:
The empty Workspaces view should teach the actual create-then-launch workflow instead of showing generic selection copy or fabricated launch controls. This PR makes the main pane explain that workspaces are PR or issue worktrees, shows the real disabled launch surface populated from configured/detected targets, and keeps the Create Workspace action inline with the prose users will see before they have any workspaces.

It also makes the Create Workspace hover guidance available through accessible descriptions in PR and issue detail views, and adds a repo-local context-sync entry point so the stop hook no longer depends on a missing global command.

Validated locally with targeted frontend unit coverage, settings server e2e coverage, the full-stack empty Workspaces Playwright case, frontend package checks, roborev fix closure, and the repo-local context-sync check.

Screenshot:

### DIFF
--- a/frontend/openapi/openapi.yaml
+++ b/frontend/openapi/openapi.yaml
@@ -5641,6 +5641,12 @@ components:
           type: array
         fleet:
           $ref: "#/components/schemas/FleetSettingsResponse"
+        launch_targets:
+          items:
+            $ref: "#/components/schemas/LaunchTarget"
+          type:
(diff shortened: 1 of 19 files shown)


mariusvniekerk:
Empty Workspaces onboarding screenshot:

PR #580, opened 2026-06-23
docs: define repo source browser

mariusvniekerk:
middleman already has local clones and provider-aware repository inventory, but it did not have a shared contract for reading repository source without turning each forge into a separate contents browser. This PR pins the stack contract first: read-only, local-clone-backed, provider-aware, bounded file reads, explicit ref freshness semantics, Markdown asset safety, and reviewable implementation slices.

Reviewers should treat this as the source-browser contract for the rest of the stack. Later PRs should either follow it or call out deliberate deviations in their own descriptions.

PR #581, opened 2026-06-23
feat: add repo source browser API

mariusvniekerk:
The UI needs a repo-code API that preserves provider identity all the way to clone and cache selection; owner/name route placeholders are not enough for nested repos, self-hosted hosts, or default-host routing. This PR adds the read-only backend foundation so later UI branches can depend on generated, bounded contracts instead of inventing filesystem access in the frontend.

The API keeps risky behavior out of request hot paths: reads come from middleman-owned local clones, refs resolve with explicit stale-token metadata, large/binary/unsupported assets become typed states, and tag pruning is left to separate maintenance rather than source-browser refreshes.


mariusvniekerk:

- #580
    - #581 ◀
        - #582
            - #583
                - #584

PR #582, opened 2026-06-23
feat: add repo browser UI state boundary

mariusvniekerk:
The source browser needs a stable frontend state boundary before the full page is assembled. This PR establishes typed route, store, and shared file-tree behavior so repo browsing can reuse middleman's provider-aware API contracts without coupling source browsing to PR diff state.

That boundary matters because the final view has to coordinate ref changes, path selection, last-changed metadata, and file-category filtering across multiple entry points. Keeping those rules in shared UI state prevents each caller from rebuilding its own partial browser model.


mariusvniekerk:

- #580
    - #581
        - #582 ◀
            - #583
                - #584

PR #583, opened 2026-06-23
feat: add repo browser source view

mariusvniekerk:
With the API and store boundary in place, maintainers need an actual code-reading workspace rather than another summary-card drilldown. This PR assembles the source view around ref and path navigation, the reusable file tree, source and Markdown rendering, and selected-file commit history.

Repository content is untrusted and refs can move, so the view keeps raw bytes and asset decisions behind backend preflight contracts while still giving reviewers a usable, forge-like repository browser.


mariusvniekerk:

- #580
    - #581
        - #582
            - #583 ◀
                - #584

PR #584, opened 2026-06-23
feat: add repo browser entry points

mariusvniekerk:
Repo browsing is only useful if it is reachable from the contexts maintainers already work in. This PR adds repository-card and command-palette entry points, resolving selected PR, issue, activity, repository, and workspace context into the same provider-aware browser route when the repository is unambiguous.

The entry points make source lookup part of triage without weakening the routing model: contextual branches are used only when they resolve in the fetched clone, and uncertain workspace or fork contexts fall back to safe defaults instead of inventing local-only refs.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -20,6 +20,7 @@
         "@codemirror/view": "^6.43.0",
         "@lucide/svelte": "1.17.0",
         "@middleman/ui": "workspace:*",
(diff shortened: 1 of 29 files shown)


mariusvniekerk:

- #580
    - #581
        - #582
            - #583
                - #584 ◀

PR #591, opened 2026-06-25
Highlight repo browser YAML files

mariusvniekerk:
- Repo browser source view now treats `.yml` and `.yaml` files as YAML for syntax highlighting.
- Other file extensions continue to use Pierre's normal filename inference.


mariusvniekerk:
Not actually a bug.

PR #595, opened 2026-06-25
fix: reduce active PR refresh budget spend

wesm:
Active PR refreshes need to stay quick for PRs that are likely to be changing, but the prior active-window behavior put every recently active open PR on the same fast lane. That burned GitHub and middleman eager-refresh budget on rows that had been quiet for hours.

This PR makes the active refresh cadence depend on last activity age: hot PRs stay on the configured fast interval, older rows inside the active window fall back to a slower cadence, and never-detail-fetched PRs are still picked up immediately. It also routes GitHub active PR refreshes through the persisted pull-request ETag path so unchanged PRs can return as not modified without spending the full comments/timeline/CI refresh budget.

The important reviewer consequence is that active PRs should still show updates quickly when GitHub reports a change, while unchanged or stale-active rows stop dominating the eager-refresh budget.

### DIFF
--- a/context/github-sync-invariants.md
+++ b/context/github-sync-invariants.md
@@ -68,11 +68,15 @@ For pull requests, that means:
   a slower cadence so the Activity view stays fresh without spending the same
   request rate on hours-old rows. A missing `detail_fetched_at` remains due
   immediately (`internal/github/sync.go::activeMRDueForFastSync`).
-- GitHub detail ETags reduce the work for unchanged PRs, but a `304 Not
-  Modified` still consumes a provider request. Do not use the ETag path as the
-  only rate-limit mitigation for broad background refresh loops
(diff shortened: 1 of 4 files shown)


context/github-sync-invariants.md
```diff
@@ -61,6 +61,22 @@ For pull requests, that means:
   persisted activity, not just one subset of the detail payload.
 - Background sync cooldowns are allowed, but user-initiated refreshes must still
   be able to promote a stronger sync intent over an in-flight background fetch.
+- Recently active open PRs in the fast-sync lane are cadence-gated by activity
```
mariusvniekerk, line 64: NICE THIS WORKED!

mariusvniekerk (approved):

PR #598, opened 2026-06-25
feat: launch Kata task workspaces

mariusvniekerk:
Kata tasks need the same workspace launch path as provider issues without treating Kata issue IDs as provider issue numbers. This keeps Kata owned by the daemon while still resolving a watched repository when the mapping is clear.

This adds conservative Kata project-to-repo resolution, daemon-scoped string workspace identity, live Kata task details inside workspace sidebars, and diff controls that only show merge target comparison when a real associated PR exists.

Validation: targeted Go and Svelte tests, ui-package-check, context-sync, and embedded-browser smoke.


mariusvniekerk:
> internal/db/db_test.go:96
Several DB migration/repair regression tests were deleted while adding a new workspace schema migration, including coverage for legacy timestamp repair, broken workspace migration repair, runtime-session rename, fleet schema repair, and workspace uniqueness migration behavior.
Fix: Restore those migration tests with unaliased assert helpers instead of removing them.

WILL NOT DO.  THESE TEST ARE A CANCER AND PROMOTE BAD DESIGN PATTERNS.

wesm:
lol

PR #615, opened 2026-06-29
Fix Kata label editing affordance

mariusvniekerk:
Fixes the Kata task detail label affordance so labels stay passive during normal scanning.

- Keeps label text as passive metadata.
- Shows remove controls only after `Edit labels`.
- Renders edit-mode labels as a segmented pill with the X attached to the label token.
- Preserves adding labels as a separate action.

Passive labels:

Editing labels:


mariusvniekerk:
Passive labels:


Editing labels:

PR #620, opened 2026-06-29
Show all CI jobs in PR detail

wesm:
The PR detail CI dropdown is meant to be the complete job view for a pull request, but the previous passed/skipped cap hid part of that list behind a secondary toggle. That made the dropdown less useful for scanning a run and added expansion state that could be reset by background refreshes.

This removes the row cap so reviewers can scroll one complete CI list directly in the dropdown, while keeping the existing bucket ordering and compact chip summary.


cpcloud:
screenshots LGTM

PR #621, opened 2026-06-30
Add kata reachable graph view

mariusvniekerk:
- Adds a Svelte Flow reachable graph view for Kata tasks from task rows and detail actions.
- Keeps graph node clicks wired into the existing task detail pane and browser history.
- Adds graph controls for depth, context emphasis, layout, direction, and done filtering through one shared filter menu.


mariusvniekerk:
https://github.com/kenn-io/kata/pull/147 landed so we should want for that version to be released

wesm:
ok, i'm cutting a release there now

PR #624, opened 2026-06-30
Add require_auth login UX: auth CLI, login overlay, and /auth/logout

wesm:
## What this adds

Makes `[api].require_auth` usable from a browser without a second auth protocol. `require_auth` stays default-off; loopback dev is unaffected.

- **`middleman auth` CLI** (`cmd/middleman/auth_cli.go`): `auth url` prints the `?auth_token=` browser login link (loopback URL from runtime metadata by default, or `--base-url <url>` for a proxied address); `auth token` prints the raw bearer for scripts; `auth rotate` regenerates the token and refuses while the daemon is running — it caches the token at startup — unless `--force`.
- **`runtimelock.RotateAuthToken`**: factors the mint+write path out of `EnsureAuthToken` into a shared helper and adds offline rotation.
- **`GET /auth/logout`** (`internal/server`): expires the HttpOnly `middleman_auth` cookie and redirects to the base path. Wired above the `require_auth` gate so it works whether or not auth is enabled (the frontend cannot clear an HttpOnly cookie itself).
- **Frontend login flow**: an optimistic auth store, a global 401 interceptor at the `createRuntimeClient` chokepoint that flips it, a `LoginOverlay` that gates the SPA and funnels a pasted token through the existing `?auth_token=` cookie bootstrap, and a Settings -> Session -> Log out link.
- **Docs**: a README Authentication section.

Reuses the existing minted token (`<data_dir>/auth_token`, mode 0600) and the existing `?auth_token=` cookie bootstrap. The gated route set (`/api/`, `/ws/`) and cookie attributes (HttpOnly, SameSite=Lax) are unchanged.

## Notes for reviewers

- The 401 interceptor is composed into the app's `createRuntimeClient`. The separate roborev and diff clients in `packages/ui` build their own fetch chains and are not wrapped, so a 401 from only those does not flip the overlay; the load-bearing client (settings/boot, and the next call after any 401) is covered. Wrapping those would mean threading an unauthorized callback across the `packages/ui` boundary.
- Pre-existing on `main`, unrelated to this branch: the `WorkspaceTerminalView` delete-button unit test is red (that file is untouched here).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -174,6 +174,7 @@ Coverage of real behavior is non-negotiable; the lane is chosen by the behavior 
 - For database schema changes, follow `context/db-migrations.md`; `internal/db/migrations/` is the source of truth for schema evolution.
 - For HTTP API error envelopes and frontend error branching, follow `context/error-handling.md`; branch on stable codes/details rather than prose.
 - For retries, backoff, and single-flight dedup against flaky upstreams, follow `context/retries-and-backoffs.md`.
+- For the `require_auth` API gate, login nonces, session cookies, and token rotation, follow `context/auth-invariants.md`; the long-lived token never goes in a URL.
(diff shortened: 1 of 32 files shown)


wesm:
abandoning

PR #627, opened 2026-07-01
fix: keep pierre diffs visible while the shared highlight pool is busy

mariusvniekerk:
Scrolling a large PR diff could intermittently show nothing where a file's content should be. Every virtualizer-driven Pierre re-render fires `onPostRender`, and the wrapper's readiness gate re-evaluated visibility each time using pool-global syntax-highlight worker stats. A rendered window with no highlighted span yet — a freshly scrolled-into region, a plain-text file, or an evicted highlight-cache entry — combined with any *other* file's queued highlight task flipped the already-visible component back to `visibility: hidden`. The loading spinner is centered over the whole file shell, so on a viewport-tall file it sits offscreen and the user just sees blank space until the pool drains.

- The `rendered` flag now latches per render attempt: once content has been shown, later post-renders cannot hide it. Attempt boundaries (file change, view-mode change, full-context re-render) still reset the latch, so initial loading behavior is unchanged.
- Diffs that render without styled spans — `lang: text`, or massive diffs beyond Pierre's tokenize cap — are treated as ready immediately; waiting for pool idleness can never make them more ready.

Both regression tests fail against the unfixed component. This does not address a second, scroll-position-deterministic blanking variant (a band toggling with 1px scrolls) that points at Pierre's virtual window math and needs separate instrumentation.

### DIFF
--- a/packages/ui/src/components/diff/PierreFileDiff.svelte
+++ b/packages/ui/src/components/diff/PierreFileDiff.svelte
@@ -904,9 +904,11 @@
     return !syntaxHighlightWorkerHasPendingWork();
   }
 
-  // Mirrors Pierre's isDiffPlainText/isDiffMassive checks: these diffs render
-  // without styled spans, so waiting on worker-pool idleness would hide them
-  // for as long as any other file still has highlight tasks queued.
+  // Mirrors Pierre's isDiffPlainText/isDiffMassive checks, which @pierre/diffs
+  // does not export as of 1.2.11: these diffs render without styled spans, so
+  // waiting on worker-pool idleness would hide them for as long as any other
+  // file still has highlight tasks queued. Recheck this mirror (and whether
+  // upstream now exports the helpers) when bumping @pierre/diffs.
   function diffRendersPlainText(): boolean {
     const fileDiff = fullContextFileDiff ?? pierreFile;
     if (!fileDiff) return false;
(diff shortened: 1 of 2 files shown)


mariusvniekerk:
Re the high finding (`PierreFileDiff.test.ts:500`): I looked at what a full-stack/browser test of this fix would actually assert, and the demanded form cannot be made deterministic. Declining to add it, with evidence:

**The negative path is not reachable deterministically from outside the app.** The regression state requires rendered diff content *plus* a busy worker pool *plus* no highlighted span in the rendered window. In `@pierre/diffs` 1.2.11, `WorkerPoolManager.initialize()` resolves its highlighter via `Promise.all([getSharedHighlighter(...), initializeWorkers(...)])` — the same gate that feeds `getPlainDiffAST`. The only external control a Playwright test has over pool busyness is stalling the worker scripts, and doing that means `renderDiff` returns nothing and no diff content renders at all, so there is no forceable state with visible content and a busy pool. Once workers initialize, busyness is machine-speed timing: an "does not return to loading during scroll" assertion passes on an *unfixed* build on any fast machine (vacuous) and flakes on slow CI. A test that cannot fail on the unfixed code protects nothing.

**The behavior is wholly frontend-component-owned.** The `rendered` flag is component state toggling a CSS class; no backend persistence, normalization, or wire shape participates. Per the repo's testing lanes, the deterministic seam is the injected pool stats — which is exactly what the two new component tests use. Both were verified non-tautological: with the two fixed lines reverted, exactly those two tests fail and the other ten pass.

**The positive full-stack path is already covered.** `frontend/tests/e2e-full/diff-highlight-screenshot.spec.ts` opts into real syntax highlighting via `__middlemanForceSyntaxHighlight`, exercises the real worker pool against the real HTTP fixture path, and asserts real highlighted Pierre rows render in a real browser.

Making the negative path reachable at the full-stack layer would require adding a test-only pool-stats override to production code, which is a worse trade than the residual gap (the component test drives `onPostRender` directly, matching the contract `FileDiff.render` implements).

PR #636, opened 2026-07-05
Fix stable header tabs across modes

mariusvniekerk:
- Keep the app header tab row stable when switching between provider-backed tabs and Kata, Docs, or Messages.
- Leave the repo selector non-interactive on non-provider modes and still honor embedded `hideRepoSelector`.
- Record the TopBar spacing invariant in UI context so future header changes preserve it.


mariusvniekerk:
Header tab alignment contact sheet:

PR #638, opened 2026-07-06
Fix Kata task table gutters

mariusvniekerk:
The Kata task list now paints table chrome to the pane edges, so opening the detail split no longer leaves visible gutters around the header and selected row.

- Keep the table scroll plane flush with the pane while preserving internal row spacing.
- Add browser geometry coverage for the edge alignment.

Visual:


mariusvniekerk:
Visual capture:

PR #651, opened 2026-07-09
Fix Kata project filter scoping

mariusvniekerk:
Kata project filters could show a selected project while the task list still came from the broader backlog when no search text was entered. This PR keeps that path scoped to the selected Kata project and records the integration invariant for future work.

- Project filters stay scoped when the search box is empty.
- Task rows missing project identity are normalized from the selected Kata project before client-side filters run.

Validated with targeted Kata API/store/workspace unit tests and `make frontend-check-no-deps`.


mariusvniekerk:
Re the roborev High on `frontend/src/lib/api/kata/taskClient.ts:265` (mixed-daemon bootstrap during explicit switches): the finding was accurate for the reviewed commit `6387b3f`, but it is already resolved on this branch.

`2dc23f7a` ("fix: keep explicit Kata switches on one daemon") implements the review's first suggested fix: `switchKataDaemon` now calls `store.clearDaemonState(previousView)` before `setActiveKataDaemon(id)` and the bootstrap, and `clearDaemonState` clears both the store's `daemonId` and the client workflow binding via `bindWorkflowDaemon()` with no argument. Every bootstrap catalog/instance read therefore resolves from the newly activated daemon, the failure-restore path re-clears the binding before restoring the previous daemon, and quick capture's `resolveOperationDaemonId(undefined, true)` can only observe a binding re-established from the new daemon's accepted results. Follow-up commits `c17ed9ff` and `30822556` harden the same boundary, and `2dc23f7a` includes store, component, and e2e coverage for it.

PR #656, opened 2026-07-10
fix: standardize transient error flashes

mariusvniekerk:
- Route transient action failures through the shared danger flash across provider, Kata, Docs, Messages, settings, repository, and workspace surfaces.
- Keep validation, load, conflict, recovery, and degraded-state errors inline without duplicate action messages.
- Keep flashes visible below measured shell chrome, at the edge of headerless shells, and above retryable modal workflows.

Failed workspace-agent settings save shown in the shared danger flash


mariusvniekerk:

PR #664, opened 2026-07-13
Clean up Kata project navigation and reassignment

mariusvniekerk:
- Use the shared grouped sidebar with collapsible project areas and navigation-only project rows.
- Make issue project breadcrumbs passive and move reassignment into the More actions menu.
- Keep failed moves retryable while preventing stale completions and errors from affecting another issue.
- Replace obsolete rename workflows with configured-daemon move and keyboard coverage.


mariusvniekerk:
Addressed the pending-move race from the review. Move guards now live in the long-lived Kata workspace hosts, keyed by source issue UID, so navigating A → B → A cannot submit a second move before the original request settles. Added regression coverage for that production lifecycle.

PR #671, opened 2026-07-14
Add a resumable historical activity archive

wesm:
Middleman can now build and maintain a complete local history of issues, pull/merge requests, comments, submitted reviews, and inline review threads, then produce deterministic reports without provider requests. This uses a dedicated archive workflow because normal sync keeps a maintainer's active working set fresh but cannot prove that every historical item and dataset has been covered.

## Why the existing sync path is insufficient

Index sync prioritizes open items and recent updates; its historical backfill had no precise resume cursor, durable coverage record, or finished state. Detail refresh loads one known item's timeline in a single pass and cannot checkpoint partial progress across budget windows or daemon restarts. Existing issue and pull-request rows therefore mean "seen by sync," not "fully archived."

A reportable archive needs durable work state rather than completeness inferred from content rows.

## What this adds

- Exhaustive, bounded provider enumeration with opaque cursors, so large repositories resume after rate limits, failures, and restarts without beginning again.
- Per-repository inventory state and per-item dataset state for comments, reviews, and inline threads, including explicit pending, complete, failed, and unsupported outcomes. Archived content still uses the same domain tables as normal sync; the new tables track work and coverage rather than duplicate data.
- Provider-neutral historical readers with honest capability gaps. Unsupported datasets produce partial coverage instead of silently appearing complete.
- Provider-host budget sharing and priority admission, so archive traffic yields to index, notification, and active-detail sync.
- `middleman archive start|pause|status|report`, a status/report API, and deterministic Markdown or JSON reports over UTC ranges.
- Durable handling for removed, transferred, and inaccessible upstream items: collection stops without discarding already mirrored local history.

## Why the diff is large

Historical collection crosses provider readers, scheduling, SQLite persistence, API generation, and the CLI. More than half of the added lines are verification or generated artifacts: about 45% are tests and 8% are generated OpenAPI/client files. The user-facing commands and reports sit on top of the resumable collection and completeness machinery that makes their results trustworthy.

The branch also documents archive scope, lifecycle, coverage, maintenance, and reporting behavior in the user and maintainer documentation.


wesm:
**e2e failure root cause** (both `pr-timeline-filters.spec.ts` failures):

The failing assertions depend on review events that the fixtures seed directly into SQLite (`w1-review-bob-1` and `w1-review-force-push-same-timestamp` in `internal/testutil/fixtures.go`) but that the fake GitHub client never returns — `buildGHReview` only serves review IDs 5011-5014, with no bodies and no COMMENTED state.

This branch makes submitted reviews a complete-replacement dataset that mirrors provider truth. Opening the PR detail view fires `POST .../pulls/github/acme/widgets/1/sync/async`, and that refresh replaces the review family with the fake provider's set, deleting the two DB-only seeded reviews. Reproduced locally against `cmd/e2e-server`: before the sync the API returns all six review events; after it, only `review-5011..5014` remain — no "COMMENTED" row and no "Same timestamp reviewer note between force-push IDs." body, which is exactly what both assertions look for.

So the server is behaving as designed (refresh mirrors provider truth, deleting rows the provider no longer returns — the same reconciliation maintenance would apply slowly, surfaced immediately by the detail-view sync). The fixtures predate that model: they assume direct event seeds survive a refresh.

Suggested fix: serve the narrative reviews from the fake provider (add them to the FixtureClient review list, extending `buildGHReview` to carry a body) instead of seeding them only in the DB, so provider truth and the seeded timeline agree. While in there, note the fixtures already disagree with the fake provider about review 5014 (seeded APPROVED at +5h vs provider DISMISSED at +6h); after replacement the provider version wins, so the direct seed is dead weight.

wesm:
## Why this needed new machinery instead of riding the existing sync path

middleman's sync was built to keep a maintainer's *working set* fresh, and its shape reflects that:

- Index sync enumerates open items and recent updates. Closed/historical items were only touched by an incidental backfill that fetched up to two pages of closed items per cycle, with no precise resume point, no record of what had been covered, and no notion of being finished.
- Detail refresh drains one item's full timeline in a single pass, and only for items already in view. It can't checkpoint partway through an item, and nothing records whether an item's comments, reviews, or inline threads were ever fetched at all.
- A row in the issues/PR tables means "sync saw this at some point", not "its history is complete". Nothing in the schema could answer "is this repo fully archived?" or "which datasets are missing and why".

The archive use case — a complete, reportable history of every issue, PR, comment, review, and inline thread across all configured repos and providers — needs exactly the properties that path lacks:

1. **Exhaustive, resumable enumeration.** Backfilling a repo with thousands of items spans many budget windows and daemon restarts. The new reader interface returns one bounded page per call with an opaque resumable cursor, so an interruption loses at most one page. Sync's listing has no cursor contract to resume from.
2. **Durable completeness accounting.** Two state tables track per-repo inventory cursors and per-item dataset status (comments / reviews / inline threads: pending, complete, failed, unsupported). That is what lets status honestly report *current* vs *partial* instead of guessing from row counts. The content itself still lives in the same tables normal sync writes — the archive is a work-tracking layer over existing data, not a second copy of it.
3. **Provider parity with honest gaps.** Historical enumeration of comments and reviews for GitLab and Forgejo/Gitea didn't exist anywhere in the codebase (the GraphQL bulk path is GitHub-only), and some datasets can't exist at all — GitLab has no historical review submissions — which must surface as explicit partial coverage rather than silently empty data.
4. **Budget subordination.** Archive traffic shares each host's rate budget and always yields to interactive sync, so exhaustion has to be a durable, resumable pause rather than a failure.

The user-visible surface on top of that (status/report API, CLI, report rendering) is roughly 10% of the branch.

Worth saying plainly: the first iteration also built ordered-ingestion epochs, guarded snapshot commits, removed-item shells, and a rolling reconciliation sweep on the theory that archive and sync writes would conflict. Review concluded the plain last-write-wins upserts main already uses are sufficient, and all of that machinery has been removed. The branch now stands at ~17k insertions, of which about 38% is tests and 11% is generated OpenAPI artifacts.

wesm:
rebasing

wesm:
Addressed the combined review in 02afda2a: hydration now persists bounded per-dataset pages and resumes from durable cursors before atomic publication; initial maintenance starts at the backfill start boundary; empty promotions complete immediately; configuration removal pauses and exact re-add resumes archive state; stale parent snapshots cannot replace labels; and Markdown output strips terminal control characters. Added rollback, limit, retry-resume, lifecycle, watermark, stale-label, and sanitization regressions.

wesm:
Addressed the combined review in `0c7fcfbe`:

- Deferred authentication and repository-blocked states are excluded before hydration claims are built.
- Archive inventory upserts preserve previously fetched merge-request detail fields while updating summary fields.
- Issue parent and label persistence is atomic and exposes monotonic acceptance, so rejected snapshots cannot replace labels or drive child writes.
- GitHub archive REST and GraphQL errors now retain typed authentication/rate-limit classifications and reset timestamps.
- Provider-controlled report fields are escaped as literal Markdown, including raw HTML, forged headings, links, and comments.

The accepted-snapshot regression also exposed GitHub issues with a missing `updated_at`; those now use `created_at` as their deterministic freshness boundary. The focused regressions and the repository short suite pass.

wesm:
Addressed the latest combined review and the exact-head follow-up in 10cadf77:\n\n- archive hydration now defers while higher-priority notification or active-detail work is running for the same provider host\n- merge-request inventory preserves detailed fields, and normal-sync MR parent/label writes are atomic\n- stale issue/MR children cannot publish after the live domain snapshot advances; the staged page remains retryable\n- a winning newer snapshot resets every supported dataset, including mixed complete/pending states\n- GitHub pull-request transfers are classified as moved and schedule destination maintenance\n- generated-client/SQLite tests cover transfer lifecycle and both sides of the post-acceptance publication race\n\nValidation: focused archive/database/GitHub/API packages, make test-short (4918 tests, 142 skipped), make vet, make lint, testify checks, and context-sync all passed.

wesm:
Addressed the exact-head follow-up in 197ef05d:\n\n- hydration now uses mirrored_provider_updated_at as the refresh boundary, so every supported dataset is fetched when a newer parent snapshot wins even if the claimed item carried stale complete statuses\n- provider-host priority registration now sits in the shared repository, MR, and issue sync execution paths; manual/API, periodic, watched, and item-number entry points all defer archive work consistently\n- generated-client/SQLite coverage exercises a newer MR with mixed complete/pending datasets, and real sync entry-point tests block manual repo, MR detail, and issue detail provider calls while checking admission\n\nValidation: make test-short (4923 tests, 142 skipped), make vet, make lint, focused archive/database/GitHub/API packages, testify checks, and context-sync passed.

wesm:
Addressed the latest exact-head failure-path finding in b968e662. Supported datasets are now durably reset against mirrored_provider_updated_at, and the reconciled status set is carried into failure recording. The new generated-client/SQLite regression proves an originally pending dataset can complete, a forced stale dataset can fail and become retryable, and the next attempt can refresh all supported datasets successfully.\n\nValidation: make test-short (4924 tests, 142 skipped), make vet, make lint, focused archive/database/API tests, testify checks, and context-sync passed.

wesm:
closed

PR #673, opened 2026-07-14
Docs: pull from git

mariusvniekerk:
The docs workspace could commit and push markdown edits but had no way to bring down remote changes, so a folder synced from another machine went stale until the maintainer dropped to a terminal. Pull is fast-forward only: divergence needs a merge or rebase the docs UI cannot host, so it surfaces as a clear error pointing at a real git client instead of leaving conflict markers in notes.

Along the way, the docs git plumbing moved onto kit's `gitcmd.Runner` like the rest of the codebase, replacing hand-rolled env stripping and command construction; an earlier draft that mirrored publish's full safety pipeline onto pull (fetch-URL classification, hardened upload-pack) was dropped as overwrought for the folder's own configured upstream — the same remote publish already pushes to.

- Pull-from-git button in the docs workspace header, next to commit & push, shown only for git-backed folders
- Fast-forward-only: diverged branches and dirty overwrites are refused with clear messages
- After a pull the tree, git status, and the open document refresh; the notice reports "Already up to date." or the new short commit
- The button is disabled while the editor is open so a pull can never discard an unsaved draft
- Pull and publish share a per-repository lock keyed by folder path, covering aliased folder IDs


mariusvniekerk:
Responding to the roborev review of `67bb65b` — all three findings are addressed by intent comments in `3ad2ec1` rather than code hardening, deliberately:

**High — pull bypasses the untrusted-repo config gates.** Intentional trust model, now stated in `GitPull`'s doc comment. A docs folder and its upstream are registered by hand by the maintainer; clicking pull carries exactly the trust of typing `git pull` in that directory. Publish's gates exist because status/changes previews rehash worktree content implicitly whenever the UI shows a folder — pull never runs implicitly. The shared runner still neutralizes hooks, fsmonitor, the transport allowlist, and inherited env on every pull command.

**High — pull/publish lock doesn't cover file saves.** Git's own fast-forward checkout refuses to overwrite tracked files whose worktree content differs from HEAD, so a save racing a pull gets the same protection it would against a terminal `git pull` — a refused merge, not a silent overwrite. The UI additionally disables pull while the editor is open. The lock's deliberately git-vs-git scope is now documented at the acquisition site.

**Medium — source-only fetch leaves `origin/main` stale.** Factually incorrect: git opportunistically updates the remote-tracking ref whenever the command-line ref matches the configured fetch refspec (verified empirically: `git fetch origin refs/heads/main` in a scratch clone advanced `origin/main`). The fast-forward integration test now asserts `origin/main` advances with `HEAD`, locking that in as a regression check.

mariusvniekerk:
Responding to the roborev review of `00e6599`:

- **High (pull not serialized with file saves)** — rejected per maintainer ruling on this PR. The pull/publish lock is deliberately scoped to git-vs-git operations only; file saves are intentionally not serialized because git itself refuses a fast-forward that would overwrite locally modified tracked files (`TestGitPullRefusesOverwritingDirtyWorktree` proves the dirty edit survives the refused pull against a real repository). A save landing mid-merge has exactly the semantics of saving in an editor while running `git pull` in a terminal, which is the accepted bar for this feature; the lock-scope comment in `docs_routes.go` documents this intent.
- **Medium (folder guard only before refreshes)** — fixed in `599ec17`: `pullFromGit` re-checks the initiating folder after every awaited refresh and before the notice, with a regression test that switches folders while the tree refresh is parked and asserts the stale reloads are abandoned.
- **Medium (missing `download` in optimizeDeps)** — fixed in `da64957`: the icon is registered and the documented used-vs-listed regeneration diff is empty again.

mariusvniekerk:
Responding to the roborev review of `da64957` — both findings are rejected as designs, with the intent now stated in code (`43119d5`) rather than guarded against:

- **High (ignored untracked files can be replaced by a fast-forward)** — this is git's own `pull --ff-only` semantics, and terminal-pull parity is this feature's accepted bar (maintainer ruling on this PR: pull carries exactly the trust of the user typing `git pull` in a folder they registered by hand). The `GitPull` doc comment previously overclaimed on this point; it now states that overwrite behavior is exactly a terminal pull's — tracked dirty files refused, gitignored collisions replaced — and that guarding beyond git's semantics is out of scope.
- **Medium (StripEnv drops GIT_SSH_COMMAND / GIT_AUTHOR_*)** — intentional posture change from the kit port that replaced the hand-rolled env filter this branch deleted. Global and system gitconfig stay readable, so the supported customization surface is gitconfig: `core.sshCommand`, `user.name`/`user.email`, `credential.helper` all work. Reintroducing a selective GIT_* allowlist would resurrect exactly the hand-maintained classification the port removed. The runner comment now documents this decision.

mariusvniekerk:
The Security reviewer is being absurd. merging.

PR #685, opened 2026-07-15
fix: stop endlessly re-syncing workspace PRs whose local tracking ref is stale

mariusvniekerk:
A PR shown in the workspace view was refreshed continuously, restarting as soon as each refresh finished. Two independent loops caused it.

**Pushed-head sync loop.** The pushed-head observer treats any mismatch between a workspace's local remote-tracking SHA and the provider head SHA as "the provider lags a local push" and re-enqueued a full PR provider sync every retry interval. When the PR advanced from another checkout, the local tracking ref is the stale side and never converges, so the same PR was re-synced forever, each completion making the frontend reload the visible PR detail. A successful sync is an authoritative provider answer: if the head still differs afterwards, another sync cannot converge, so the observer now stops re-enqueueing until the local tracking ref moves. Refresh timestamps reset when the observed SHA changes so a dropped enqueue cannot suppress a later push; failed refreshes keep the retry interval.

**Workspace-status broadcast loop.** Every enrichment completion and every tmux prune pass broadcast `workspace_status` unconditionally (~40 events/min observed live with one open workspace). A busy agent moves the tmux-activity timestamp on every 5s probe, so each broadcast made clients refetch and each refetch scheduled the next enrichment past the 5s TTL. Completions now notify only for durable changes (first completion, divergence movement, error-state change) and prune broadcasts only when it pruned; activity ordering still flows through the workspace list's own polling.

- Workspace-view PRs with stale local tracking refs are no longer re-synced and re-rendered in a loop
- Open workspace views no longer refetch every few seconds while idle or while an agent is streaming output
- Push-triggered refresh, failure retries, and the pending-to-fresh enrichment transition behave as before
- Convergence and broadcast rules are recorded in `context/workspace-apis.md`

### DIFF
--- a/context/workspace-apis.md
+++ b/context/workspace-apis.md
@@ -200,6 +200,17 @@ Test fixtures that seed PR rows must either carry a same-repo
 unknown-provenance setup resolves heads exclusively through the fork-safe ref
 and fails outright when the remote does not serve it.
 
+## Pushed-Head Refresh Convergence
+
+A tracking-ref/provider-head mismatch is not proof of a local push: when the
+PR advanced from another checkout, the local tracking ref is the stale side
+and a provider sync can never converge. The observer therefore enqueues a PR
+sync on mismatch, retries on failure after `pushedHeadRefreshRetryInterval`,
+but must stop once a refresh for the same observed SHA succeeded and the
+provider head still differs (`LastRefreshSucceededAt >= LastRefreshEnqueuedAt`)
+— otherwise the visible PR is re-synced and re-rendered forever. A tracking-ref
+move restarts the cycle.
+
 ## Sidebar Ordering
 
 The workspace sidebar has two separate activity concepts:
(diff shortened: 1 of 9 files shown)


mariusvniekerk:
Pushed two follow-ups after probing the live instance:

- The live refresh loop turned out to be a second mechanism: enrichment completions and tmux-prune passes broadcast `workspace_status` unconditionally (~40/min observed on the events stream with one open workspace), and each client refetch scheduled the next enrichment past the 5s TTL. Broadcasts are now gated on durable changes (first completion, divergence movement, error change; prune only when it pruned). Tmux-activity ordering still flows through the list's own polling.
- Both review findings were confirmed valid and are addressed: refresh timestamps reset when the observed SHA moves and completion markers no longer stamp a different SHA's cycle (with a reproduction test for the dropped-enqueue suppression), and a server e2e test now covers the non-converging refresh end to end: one sync, silence past the retry interval, refresh again on a real push.

PR #690, opened 2026-07-17
feat: support GitHub PATs selected by repository owner

mariusvniekerk:
Fine-grained GitHub PATs are commonly scoped to one organization, while middleman previously assumed one host credential could authorize every repository on that host. This draft separates owner-selected authorization from GitHub identity-scoped capacity so maintainers can monitor repositories across multiple owners without overstating rate limits.

The branch currently includes configuration-only owner mappings, stable PAT/App identity resolution, identity-scoped rate persistence and sync budgets, repository-routed API clients, owner-aware preview and managed Git authentication, reload boundaries, and operator documentation. Existing single-token, repository override, GitHub App read, host fallback, and GitHub CLI behavior remain part of the intended precedence.

This remains a draft while the final review findings around App-only snapshot routing and queued-notification rate-limit isolation are completed. The current pushed state passed repository commit and push hooks, including short tests, lint, API regeneration, testify checks, and nilaway.


mariusvniekerk:
> A GitHub and Forgejo configuration sharing a hostname can therefore attach the GitHub PAT to Forgejo clone or fetch requests.

This should be impossible

mariusvniekerk:
Both findings from the combined review of d8ac3b9 are addressed in 217b81eaa:

- `ProviderTokenSources` dedup: `ValidateSharedHostCloneTokenSources` now validates each non-GitHub repository's effective token descriptor before plan deduplication, so conflicting same-host `token_env` overrides are rejected at startup and reload (`TestValidateRejectsConflictingRepoTokenSourcesOnSharedHost`).
- `discoveryClientForOwner`: owner and fallback PAT routes now take precedence for repository discovery; the selected-App discovery client applies only when owner routing returns `MissingRouteError` (`TestRoutedClientDiscoveryPrefersOwnerAndFallbackRoutes`).

The same commit also closes the remaining open local review findings on this branch (write-identity budget/provider-work accounting, implicit `github_token_env` fallback, reload probe token cache, typed archive capability errors).

mariusvniekerk:
Both findings from the combined review of 217b81e are addressed in c69876ad — both were host-wide scopes that are now identity-scoped:

- **Notification watermarks are per repository identity.** `middleman_notification_sync_watermarks` is keyed `(platform, host, owner, name)` (rebuilt in the PR-local migration, renamed `000040_identity_scoped_sync_state`), and each repository advances its own watermark: a repo with an unavailable route or exhausted PAT reports its error while healthy siblings advance, and a newly tracked repo full-syncs without resetting the host. The `tracked_repos_key`/`sync_cursor` machinery is retired with the host key.
- **Clone credentials validate per `(provider, host)`.** Providers sharing a hostname may carry different fallback tokens; validation only rejects conflicting effective chains among repositories of one provider-host pair, and the ownerless host fallback is disabled when providers on one hostname disagree, instead of borrowing whichever provider came first. The dead host-wide `validateProviderHostKeys` copy is removed.

mariusvniekerk:
Follow-up in 12cb2f40 addressing the remaining review findings on 217b81e/c69876a:

- Owner discovery now unions the PAT route listing with selected-App discovery, deduped by repository ID, and fails closed if either configured source fails — selection-only grants and PAT-only repositories both stay visible.
- Notification sync skips a repository whose credential route cannot resolve instead of aborting the whole host; healthy siblings keep advancing their per-repo watermarks.
- A disabled ambiguous shared-host clone fallback is honored at route resolution instead of falling through to the GitHub unscoped route.
- The implicit github.com fallback is probed best-effort at startup: a valid default token keeps ownerless APIs routed, an invalid one is skipped with a warning instead of failing startup (explicit fallbacks still fail hard).
- Archive-context repository reads skip the viewer-permission overlay so archive traffic never spends the write identity.
- Context docs and the plan doc were reconciled with the per-repo watermark model and union discovery semantics.

PR #691, opened 2026-07-17
Speed up selected workspace diff loading

mariusvniekerk:
Workspace diff views were repeating expensive Git work and could keep the prior workspace's sidebar visible while a new selection loaded.

- Prepare and cache diff data only for the selected workspace.
- Keep shell and runtime readiness independent from diff preparation and browser parsing.
- Replace stale sidebar content immediately, then mount the selected diff when its runtime is ready.
- Revalidate selected diff data asynchronously on manual refresh and notify the browser when it is ready.
- Reduce the reported workspace's warm server loads to 0.25ms for files and 0.35ms for diff after a 90.8ms cold preparation.

### DIFF
--- a/context/workspace-apis.md
+++ b/context/workspace-apis.md
@@ -271,6 +271,13 @@ server check exactly.
   captured load token and generation still own the store; stale preview work
   fails without mutation
   (`packages/ui/src/stores/diff.svelte.ts::loadWorkspaceFilePreview`).
+- Preserving refreshes publish only a fresh coherent files/diff pair; stale
+  responses retain the visible pair and retry because same-fingerprint
+  validation emits no change event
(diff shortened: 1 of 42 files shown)


mariusvniekerk:
Accepted tradeoff for `67e27046`: active diff scopes intentionally expire after ten minutes without a foreground read so selected workspaces cannot pin every previously viewed base/range indefinitely. Background validation deliberately does not renew foreground access. The attempted per-scope heartbeat in `b8415d0c` introduced competing lease ownership with fleet HEAD polling, so that commit was removed rather than layering another watcher exception onto the cache. Indefinite live freshness for a visible-but-unread scope is outside this performance change; it should only be revisited with a simpler ownership model.

PR #699, opened 2026-07-19
Fix Activity PR detail live refresh

wesm:
The Activity sidebar and selected PR timeline now update from the same live invalidation while retaining the newest successful selected detail when requests finish out of order.

- Show newly persisted PR events in the open Activity timeline immediately, including after stale reconnect recovery
- Treat provider aliases and omitted default hosts as the same selected PR identity
- Keep rapid selection changes pinned to the current PR across provider and host identities
- Preserve a pending successful detail load when a newer live refresh fails


wesm:
I'm testing out this fix locally before merging it

wesm:
This appears to work fine, merging

PR #701, opened 2026-07-19
fix: restore compact PR comment spacing

wesm:
Heading-first Markdown comments were stacking three separate top-spacing rules, so automated review comments still looked detached from their headers after the shared card gap was removed.

- Make the timeline wrapper the single owner of the 8px header-to-body interval
- Apply the spacing to heading-first Markdown while preserving compact and system cards
- Add real-browser geometry coverage for automated-review-shaped comments
- Keep canonical Typeahead selection tests aligned with the updated kit-ui dependency

Corrected PR comment spacing


wesm:
this still requires work

PR #702, opened 2026-07-19
Run Linux CI on self-hosted runners

wesm:
Linux CI repeatedly pays hosted-runner setup and dependency transfer costs. Organization-managed ephemeral runners can reuse warm local state while Windows and release jobs remain hosted.

- Move Linux CI jobs to the self-hosted runner group while keeping Windows and release jobs hosted.
- Disable redundant setup-go cache transfers across CI and release workflows.
- Bound frontend worker counts and Go package, runtime, and memory concurrency for high-core, memory-bounded jobs.
- Make workspace test fixtures independent of tmux availability by using the in-process PTY owner.
- Give independently provisioned E2E states separate timeout budgets.
- Record the trusted-collaborator and approved-run boundary for public pull-request CI.

### DIFF
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,3 @@
+self-hosted-runner:
+  labels:
+    - kenn-linux-x64
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,7 +113,7 @@ jobs:
     name: Go tests
     runs-on: kenn-linux-x64
     # Bound package fan-out on high-core runners so concurrent test binaries
-    # stay within the job's memory budget.
+    # and in-package parallel tests stay within the job's memory budget.
     env:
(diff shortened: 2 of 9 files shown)


mariusvniekerk:
configwatch: new watcher: couldn't initialize inotify: too many open files , seems like come config needed for the self-hosted runners.

wesm:
Thanks — yes, this was a runner-host configuration issue. `fsnotify.NewWatcher` was receiving `EMFILE` from `inotify_init` after concurrent jobs exhausted the host’s per-user inotify instance limit; it was not the test process’s ordinary `ulimit -n`.

We’ve raised and persisted the host’s inotify instance/watch limits and added deployment verification so a runner below the required capacity fails closed. After rollout, the previously failing frontend browser, Go, and race jobs all passed in [the rerun](https://github.com/kenn-io/middleman/actions/runs/29742475609).

PR #703, opened 2026-07-20
fix: eliminate PR refresh judder and duplicate indicators

wesm:
Refresh state no longer moves PR review content or shows duplicate progress notices.

- Keep stale diff warnings clear of file controls while preserving the review layout during loading and refresh.
- Use the existing PR-detail "Syncing" indicator instead of a second "Refreshing..." banner.
- Add real-browser coverage for both refresh paths.

### DIFF
--- a/packages/ui/src/components/detail/PullDetail.svelte
+++ b/packages/ui/src/components/detail/PullDetail.svelte
@@ -2608,6 +2608,7 @@
 
   .pull-detail-content {
     container: pull-detail / inline-size;
+    position: relative;
     display: flex;
     flex-direction: column;
     gap: 16px;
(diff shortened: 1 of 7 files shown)


wesm:
this needs more work

wesm:
Working on this, still more crap broken

packages/ui/src/components/detail/PullDetail.svelte
```diff
@@ -1653,12 +1652,6 @@
               </Button>
             {/snippet}
 
-      {#if detailStore.isStaleRefreshing()}
-        <div class="refresh-banner">
-          <StatusDot status="working" label="Refreshing pull request details" size={5} />
-          <span aria-hidden="true">Refreshing...</span>
-        </div>
-      {/if}
```
wesm, line 1661: @mariusvniekerk not sure where this came from but here was the canard

mariusvniekerk (approved):

PR #711, opened 2026-07-20
fix: recover Mermaid after frontend updates

cpcloud:
An already-open Middleman tab can outlive a binary update. When that tab later encounters its first Mermaid diagram, Vite can request an obsolete content-hashed sequence-diagram chunk from the new server. The server correctly returns 404 for the stale asset, and Kit restores the escaped Mermaid source after the infrastructure failure, matching the reported plain-code rendering.

Middleman now handles Vite's `vite:preloadError` event at the application deployment boundary. It fetches the server's current HTML without cache and reloads only when that document points to a different frontend entrypoint, while transient failures continue into the existing feature-level retry paths. A source-and-target session marker prevents repeated reloads while a deployment is incomplete and is retired after loading either the target or a newer frontend, allowing later rollback and redeploy cycles to recover. The reusable Kit renderer remains unchanged.

Browser coverage exercises the reported `sequenceDiagram` beside an ordinary TypeScript fence, keeps the stale sequence chunk permanently unavailable, and requires the reload to boot a distinct updated asset graph. Focused state-transition coverage protects rollback after a reload lands on an intervening deployment, while existing coverage preserves the repeated-failure retry flow for other lazy features.

### DIFF
--- a/frontend/src/main.ts
+++ b/frontend/src/main.ts
@@ -3,9 +3,9 @@ import { pushModalFrame } from "@middleman/ui/stores/keyboard/modal-stack";
 import { mount } from "svelte";
 import App from "./App.svelte";
 import "./app.css";
+import { prepareFrontendReload, retireFrontendReload } from "./lib/utils/frontendReloadGuard.js";
 import { initMarkdownImageExpansion } from "./lib/utils/markdownImages.js";
 
-const viteReloadKey = "middleman:vite-reload";
 const currentFrontendEntrypoint = new URL(import.meta.url);
 
 function frontendEntrypoints(document: Document, baseUrl: string): URL[] {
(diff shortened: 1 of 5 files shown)


mariusvniekerk (approved):

frontend/src/main.ts
```diff
@@ -5,6 +5,45 @@ import App from "./App.svelte";
 import "./app.css";
 import { initMarkdownImageExpansion } from "./lib/utils/markdownImages.js";
 
+const viteProbeParam = "_middleman_vite_probe";
+const viteReloadKeyPrefix = "middleman:vite-reload:";
+
+function viteAssetUrl(payload: unknown): URL | null {
+  if (!(payload instanceof Error)) return null;
+  const match = payload.message.match(/https?:\/\/[^\s'")]+/);
```
cpcloud: what in god's name is happening here.

PR #723, opened 2026-07-22
Compact two-line sidebar cards for PR and issue lists

wesm:
Redesigns the PR and issue sidebar cards as a fixed two-line layout so more items fit on screen while staying readable.

- Cards drop from ~80px to ~50px tall: title line with small label name chips and the item number pinned top-right; repo, author, indicators, and relative time on the second line
- Labels render as compact name chips (first two plus a +N overflow count)
- Repo shows as plain hash-colored text rather than a pill, and only truncates when the row genuinely runs out of space (author yields first)
- Kanban status chips no longer appear on PR rows; issues show a chip only when closed
- Mobile keeps its phone sizing and 2-line title clamp with the same structure
- Shrinks the compact CI pending dot so it no longer breaks uniform row height


wesm:
Re: the roborev finding on LabelRow.svelte:47 (label names only reachable via mouse-hover title on an aria-hidden span) — resolved as of 8b638f5f, which replaces the color-only dots with visible label name chips on the title line (first two labels plus a +N overflow count). The names are now plain visible text for all users, so the tooltip/sr-only machinery was removed along with the dots. An intermediate commit (6e1ff7ac) had addressed this with a focusable kit-ui Tooltip; it was superseded when the maintainer opted for visible chips.

wesm:
Re: the roborev finding on LabelRow.svelte:42 (non-shrinking compact label row clipping the title/number at narrow widths) — confirmed and fixed in 43b76ab5. Pre-fix measurement reproduced the clip: with two long labels plus an overflow count in a 200px row, the item number's right edge landed at 271px. The compact label cluster now shrinks before the title (mirroring the meta line's shrink-priority idiom) and the number keeps flex-shrink: 0, so it never clips. Added Vitest browser-project coverage mounting the card at 200px (minimum sidebar) and 360px inside .mobile-main with two >20-char labels + overflow count, asserting the number and label cluster stay inside the row bounds and the title keeps nonzero width; the test fails against the pre-fix CSS.

PR #724, opened 2026-07-22
Inline workspace surface on Activity, PRs, and Issues

wesm:
## Summary

- The Workspaces top-bar tab restores the last workspace route (`/terminal/{id}`, fleet routes included) instead of resetting to the bare list; clicking it while already in workspace mode is a no-op.
- Selecting a PR or issue whose detail has an attached workspace now shows the live terminal in a resizable dock beneath the detail on the Activity, PRs, and Issues tabs (desktop layout only). The dock has no header bar of its own; Expand Terminal / Show Details and Collapse Terminal sit in the workspace toolbar next to Delete.
- It is the same live workspace instance everywhere: flipping between the Workspaces tab and an inline dock reparents the running terminal (sockets, scrollback, and session tabs survive). Dock height persists per surface.
- Creating a workspace from an inline-capable detail claims the dock in place instead of navigating away; the open action becomes Focus Terminal with Open in Workspaces as a secondary action. Deleting a workspace returns the detail to full height. Focus and mobile presentations keep today's navigation behavior.
- kit-ui pin advanced for the new BottomDock controlled-height API (kenn-io/kit-ui#39); the pin should move to the merge commit once that PR lands.

## Behavior notes

- After visiting a terminal, one workspace stays connected (hidden) while browsing non-workspace pages for the rest of the session, so returning is instant.
- While the hidden host is parked, the workspace list sidebar unmounts entirely: no background workspace polling on non-workspace pages.


wesm:
Inline workspace dock under an issue detail — split view (controls in the workspace toolbar next to Delete, no dock header bar):



Expanded view (title strip, terminal filling the pane):

wesm:
All four findings addressed in eb943f53..a5bd0e2e:

- **Outdated e2e navigation assertions (High)** — confirmed: all four tests failed deterministically. Creation specs now assert the issue route stays active and the workspace appears in the inline dock; the first spec keeps explicit coverage of the secondary Open in Workspaces navigation to /terminal/{id}. The same commit fixes four more specs that hit strict-mode violations on .kit-sidebar-layout__main (the parked host adds a hidden second instance on every desktop page); they now filter to the visible layout. (7b8757fa)
- **Hidden terminal global effects** — the Cmd/Ctrl+] listener and both resize-clamp effects are now gated on hostVisible plus the right sidebar rendering, so a parked host can't consume the shortcut on other pages or persist a width clamped against display:none geometry. Browser regression test fails without the fix. Contract documented in context/ui-interaction-contracts.md. (b7f66bb8)
- **Stale remembered route after deletion** — notifyWorkspaceDeleted now carries the fleet host key and forgets the remembered terminal route; the router also refuses to re-remember a forgotten route, covering deletion from the terminal page where the subsequent navigate to /workspaces re-captures the dying URL. (b19328fb)
- **Late responses on inactive detail surfaces** — both detail components track a request generation bumped per request, on identity change, and on destroy, and reject stale workspace-create responses immediately after resolution — before the branch-conflict parse, error, success, and finally handling. Unmount-mid-flight tests fail without the fix. (1c7fea98)

Validation: full unit (3244), browser (147), typecheck, and the full chromium e2e suite. One unrelated failure remains in full-suite runs: mobile-routes.spec.ts:372 ('mobile activity card routes to a focused thread detail') fails on a system-event-summary overflow at this branch's merge-base state too (reproduced with this PR's changes stashed), passes in isolation, and none of this branch's files touch that path — it appears to trace to timeline changes on main (#708) and is left for a separate fix.

Also bumped the kit-ui pin to the merged #39 squash commit (a5bd0e2e).

wesm:
All three findings addressed in a5bd0e2e..371fe937; each regression test was verified to fail against the unfixed code.

- **Route memory misses Back-exited terminal visits** — navigate() now records the destination after assigning the new route, so a terminal entered via navigate() and left via browser Back is remembered; popstate coverage added. (b1581d58)
- **Expanded mode survives panel unmount** — WorkspaceDockPanel also resets expanded to split at teardown, covering unmount paths the reset-on-inactive effect never observes; unmount regression test added. (e319a79f)
- **Delete success discarded after switching workspaces** — both delete and force-delete now report onWorkspaceDeleted for successful responses before the current-workspace guard, so claims, tombstones, and route memory update regardless of the current selection; prompts, flashes, and navigation stay gated. Tests cover both paths with the switch happening mid-flight. Contract added to context/ui-interaction-contracts.md. (371fe937)

Validation: full unit (3247), browser (148), typecheck, lint, and the full chromium e2e suite — 531/532, the one failure being the previously reported pre-existing mobile-routes overflow flake that reproduces at this branch's base.

wesm:
Both findings from the review of 371fe93 verified against the codebase and fixed.

**Deletion tombstones only active claims** — confirmed: PRListView/IssueListView release their inline claims in effect cleanup on unmount, so a delete issued from the Workspaces or terminal tab found no claim to tombstone and the cached detail envelope re-claimed the dead workspace on revisit. Fixed in 844084b: the workspace-host store keeps workspace-id → identity metadata past claim release (recorded on claim and recordCreated, dropped on deletion/test reset) and tombstones + fires invalidation listeners for the remembered identity even without an active claim. Store tests cover claim → release → delete → stale envelope for both the claimed and created-then-reconciled paths.

**Parking leaves the sidebar mounted and menus/overlays alive** — confirmed: `hideRightSidebar` was only set for embedded hosts, and the workflow preset / terminal options / launch menus had no hostVisible wiring, so Back with the font picker Modal nested in the options popover left an invisible overlay, focus trap, and modal-stack frame owning Escape. Fixed in 8bd82a5: WorkspaceHost unmounts the right sidebar while parked (open state and width persist for re-reveal), and all three toolbar menus take `hostVisible` and close when it goes false. Menus are transient, so re-revealing does not resurrect them — unlike the confirm dialogs, which restore by design. Browser tests cover the nested-modal park (popover + font Modal open → parked → no overlay, stack depth 0) and the parked right-sidebar unmount/restore cycle.

All four new regression tests were verified to fail with the fixes stashed. Validation: full unit suite (3249 passed), full browser suite (150 passed), typecheck, lint on changed files, and the full chromium e2e suite (532 passed, 4 skipped — the pre-existing mobile-routes overflow flake noted in earlier rounds did not occur this run).

wesm:
Both findings from the review of 8bd82a5 verified against the codebase and fixed.

**Expanded mode survives direct claim replacement** — confirmed: the list views' claim effect calls `claim(identityB)` directly when the new selection's detail already matches, so `claims` goes A → B in one store write. WorkspaceDockPanel's reset-on-inactive effect never observes an inactive gap and its teardown reset never runs, leaving the new item's detail hidden behind a fullscreen terminal. Fixed in f3667aa: `setClaim` resets the surface to split when it replaces a claim holding a different identity; same-identity re-asserts (ref status changes) keep an expanded dock intact. Store test covers both the replacement reset and the same-identity no-reset.

**Hidden detail keeps handling window-level commands** — confirmed: the `labels.edit` palette action dispatches the open-label-picker window event based on selection/detail match only, and PullDetail/IssueDetail hold that listener while mounted-but-hidden under an expanded dock, so the picker opened invisibly inside the hidden+inert subtree and appeared on the next collapse. Fixed in 55e84f1: since the command is an explicit detail operation, both handlers restore the surface to split before opening the picker when the dock is expanded over their claimed identity, so the picker lands visibly instead of being dropped. When the rendered detail is not the claimed identity (detail fully visible), the dock mode is left untouched. Component tests cover the command while expanded (split restored, picker opens) and the not-claimed case (mode untouched) for PullDetail, plus the expanded case for IssueDetail.

All three new regression tests were verified to fail with the fixes stashed. Validation: full unit suite (3253 passed), full browser suite (150 passed), typecheck, lint and format on changed files, and the full chromium e2e suite (531 passed, 4 skipped, 1 failure in stale-asset-recovery.spec.ts that passed on isolated rerun — the same contention flake seen in an earlier round; my diff touches no asset-reload path).

wesm:
Both findings from the review of 55e84f1 verified and addressed.

**Creation-to-inline-dock handoff not actually verified** — confirmed: `.workspace-dock-panel` is rendered unconditionally by WorkspaceDockPanel (only the BottomDock inside is gated on an active claim), so all four assertions passed pre-claim. Fixed in d74c671: the four sites now assert `.workspace-dock-slot .workspace-host-wrapper` — the slot exists only while a claim holds the surface with the dock open, and the host wrapper matches inside it only after the workspace host physically reparents into that slot, so the assertion fails if `recordCreated` never activates or hosts the dock. Added the missing full-stack PR creation-to-dock test (its own POST /workspaces endpoint) through the real HTTP API, SQLite, git worktree, and tmux, mirroring the issue test: URL stays on the PR, the created workspace claims and hosts the dock, the worktree checks out the reported head ref, and Open in Workspaces still navigates to /terminal/{id}. Writing that test exposed a fixture inconsistency it now guards: `buildGHPR`'s placeholder provider head ref ("feature") disagreed with the seeded DB row and the diff repo's real branch ("feature/caching"), so merely opening the PR page synced the placeholder into the DB and broke worktree provisioning; widgets#1's provider fixture now pins the seeded branch.

**Raw provider strings in workspace identity** — confirmed: the router keeps the path's provider segment verbatim (canonicalization was only applied for default-host lookup), so `/pulls/gh/...` produces identities keyed "gh" while store data uses "github", splitting claims, overrides, and tombstones for the same item. Fixed in a09220b: `identityKey` (workspace-host store) and `identityEquals` (workspace-inline) canonicalize the provider before keying or comparing. Alias-equivalence tests cover claims/overrides/tombstones through mixed alias and canonical forms, plus direct `identityEquals` cases; both were verified to fail with the fixes stashed.

Validation: Go `make test-short` (5127 passed) plus full `cmd/e2e-server` and `internal/workspace` runs for the fixture change, full frontend unit suite (3256 passed), browser suite (150 passed), typecheck, vet, golangci-lint, lint/format on changed files, and the full chromium e2e suite (532 passed, 4 skipped, 1 failure in stale-asset-recovery.spec.ts that passed on isolated rerun — the recurring contention flake previously flagged on this PR).

wesm:
Both findings from the review of d74c671 verified and fixed.

**WorkspaceHost mounts before appReady** — confirmed: the page-branch chain in the desktop shell gates on `appReady`, but the host sat after it ungated, so a direct /terminal/{id} load fired workspace/runtime fetches before the /healthz readiness poll and settings initialization finished. Fixed in the first commit: the host is now wrapped in `{#if appReady}` while staying outside the page branches, so it still mounts exactly once per shell and survives page switches (`appReady` only resets when the whole shell tears down). The shared API mock now lets route overrides answer readiness paths first, and a new browser app-harness test holds /healthz at 503 through more than a full poll cycle, asserts zero workspace requests during that window, then releases readiness and asserts the workspace fetch and terminal view arrive.

**Focus not restored when an open dock closes** — confirmed: only expanded→split transitions restored focus; a toolbar collapse from split, or a deletion/release while split, unmounted the BottomDock subtree and dropped focus on `<body>`. Fixed in c4ed8c2: the panel tracks dock-open state and, when an open dock closes for any reason, focuses the detail wrapper after the DOM update — guarded so it only fires when focus actually fell to `<body>`, meaning a dock closing in the background can never steal focus from an input the user is typing in. Browser tests cover toolbar collapse, split-mode release, and the no-steal case.

All three regression tests were verified to fail with the fixes stashed (the no-steal guard test passes either way by design). Validation: full unit suite (3256 passed), full browser suite (154 passed), typecheck, lint/format on changed files, and the full chromium e2e suite (532 passed, 4 skipped, 1 failure in mobile-routes.spec.ts:372 that passed on isolated rerun — the pre-existing mainline flake flagged in earlier rounds on this PR).

wesm:
Both findings from the review of c4ed8c2 verified and fixed.

**Expanded-to-split always focuses the detail wrapper** — confirmed: the claim-replacement reset added earlier triggers the expanded→split focus effect, which focused the detail unconditionally and pulled focus off the sidebar row the user had just selected. Fixed in 09e4780: all dock focus restoration now runs through one guard — reclaim only when focus fell to `<body>` or was still inside the closing terminal subtree (checked against the dock slot element). Collapse-from-expanded with focus on the terminal's own toolbar still returns focus to the detail per the collapse contract; focus owned by any outside control is preserved. Browser tests cover the selection-change-while-expanded preserve case and the focus-in-terminal collapse case.

**Deletion tombstones depend on inline-claim identity metadata** — confirmed: `workspaceIdentityById` is fed only by `claim`/`recordCreated`, so a workspace first opened in the Workspaces tab or deleted straight from the sidebar list had no entry, and its cached detail envelope could re-claim the dead workspace on a later desktop visit. Fixed in 9acee29 via the reviewer's first option: the deletion callback now carries the provider-aware identity. WorkspaceTerminalView snapshots it from the loaded envelope at request time — the force-delete path keeps the snapshot taken when the 409 arrived, since the live envelope can describe a different workspace by confirm time after an A→B switch — and WorkspaceListSidebar maps it from the deleted row. `notifyWorkspaceDeleted` tombstones the supplied identity alongside remembered and claimed ones. Store test covers deletion of a workspace that was never claimed inline; the WTV delete/force-delete tests now assert the identity rides the callback.

Three fix-dependent regression tests were verified to fail with the fixes stashed. Validation: full unit suite (3257 passed), full browser suite (156 passed), typecheck, lint/format on changed files, and the full chromium e2e suite (531 passed, 4 skipped, 2 failures that both passed on isolated rerun: the recurring mobile-routes overflow flake — which this round also passed at clean branch HEAD and with the changes applied on repeat runs — and a one-off 'Response has been disposed' race inside activity-drawer.spec.ts's own route interception).

wesm:
81d70c35 addresses the workspace-identity finding: `WorkspaceItemIdentity` now carries an `itemType` field, canonicalized across caller vocabularies ("pull"/"pr"/"pull_request", "kata_task") by a new `canonicalItemType`, mirroring the existing provider-alias handling. Both `identityEquals` and the host store's `identityKey` include it, so a PR and an issue sharing a repository and number no longer cross claims, creation overrides, or deletion tombstones. All producers populate it: PullDetail/IssueDetail item identities, PRListView/IssueListView/ActivityFeedView claim identities, and the workspace-envelope/sidebar-row deletion mappings (from `item_type`).

Regression tests (verified to fail against the unfixed code): identityEquals separates PR #7 from Issue #7 and treats the vocabulary aliases as one type; store-level tests cover claim, override, and tombstone isolation between a same-numbered PR and issue, plus alias-keyed claim/tombstone equivalence.

Validation: full Vitest suite (3420 passed), svelte-check/lint/format clean, full chromium e2e (530 passed; 3 pre-existing flaky specs — activity-drawer, mobile-routes, stale-asset-recovery — each pass in isolation).

wesm:
0be8c41b addresses the default-host normalization finding. A new `resolvedPlatformHost(provider, platformHost)` helper in `provider-routes.ts` (omitted/blank host resolves to `providerDefaultHost`) is now applied everywhere workspace identity is compared or keyed: `identityEquals`, the host store's `identityKey`, and the four detail-match guards (PRListView/IssueListView `detailMatchesSelected`, PullDetail/IssueDetail `detailMatchesIdentity`). This matches the normalization the detail store's request keying already used. An Activity selection without `platform_host` now claims against a detail payload carrying the concrete default host, and overrides/deletion tombstones no longer split across omitted-host and concrete-host keys.

Regression tests (all six verified to fail against the unfixed code): identityEquals omitted-host equivalence (and inequality for a non-default host), store-level claim/tombstone keying across omitted vs concrete default host, claim coverage in both list views for a selection without `platform_host`, and reconcile coverage in both detail components for an identity without `platform_host`.

Validation: full Vitest suite (3426 passed), svelte-check/lint/format clean, full chromium e2e (530 passed; 3 failures were parallel-run flakes — two detail-action-buttons suggestion specs and mobile-routes — all pass in isolation, and the full detail-action-buttons file passes 24/24 on rerun).

wesm:
b59f2287 addresses both findings.

Unmount-mid-create: `responseIsStale` is split into `identityLeft` (generation bump or identity mismatch — still discards everything) and `componentDestroyed`. A successful create response now always records its identity-scoped store override (`recordCreated`, which is claim-guarded in the workspace-host store) and fires the detail refetch even when the component unmounted — tab switch to Files in non-split mode, split-view toggle, breakpoint crossing, or drawer tab change all unmount the detail with the same item selected. Only local state, flash, the branch-conflict dialog, and navigation stay gated on the live component. The two earlier discard-on-unmount tests encoded exactly this contract, so they are replaced by "records the override when a layout change unmounts the detail mid-create" tests in both components (the leave-the-page case runs the same code path and remains safe: recordCreated does not claim a surface that moved on).

Stale guards: `stalePR`/`staleIssue` now compare via `canonicalProvider` and `resolvedPlatformHost`, matching the claim/reconcile guards, so an Activity URL with an omitted default host or an aliased provider no longer disables mutation actions on a current detail. Covered by "keeps mutation actions live when the identity omits the host or aliases the provider" tests in both components.

All four new tests were verified to fail against the unfixed code. Validation: full Vitest suite (3428 passed), svelte-check/lint/format clean, full chromium e2e (530 passed; 3 failures were an ECONNREFUSED infrastructure drop in 00-workspace-sidebar plus the known mobile-routes and stale-asset-recovery flakes — all three pass in isolation).

mariusvniekerk (approved):

wesm:
0dd73098 addresses both findings.

Workspace-host race: `recordCreated` no longer claims an unclaimed surface — it only refreshes a matching live claim's ref. Claims now come exclusively from the list views' claim effects, which read the recorded override reactively (`effectiveWorkspaceRef` tracks the overrides state) and confirm the identity against the live selection before claiming. Independently, `desiredKey()` now derives from the visible surface's own claim and falls back to `lastInlineKey` only while parked, so no late write to the sticky key (a stray claim, a stale terminal-route key) can rebind the hosted terminal underneath a dock displaying a different claim. Regression tests: a cross-surface late-response test (create for PR A lands after the user moved to the issues page with workspace B claimed — override recorded, issues dock keeps ws-B, prs surface stays inactive) and a sticky-key-drift test (a claimed surface keys the host even after `rememberTerminalRouteKey` moved the sticky key).

Stale refetch: the post-create refetch is skipped when the originating component was destroyed — its frozen identity always matches its own props, so it could reload the old item into the shared detail store after a different selection. The creation override is still recorded for replacement components, which load their own detail on mount. The round-10 unmount tests now also assert no refetch fires.

All four new/extended tests verified to fail against the unfixed code. Validation: full Vitest suite (3430 passed), svelte-check/lint/format clean, full chromium e2e (532 passed, 1 failure — the known activity-drawer route-interception flake, passes in isolation).

wesm:
9cc00dd6 addresses both findings.

Alias-invalidation: both detail components' route-reset effects now snapshot the item identity and skip when `identityEquals` reports no genuine change, instead of firing on any raw prop change. An alias-only re-expression of the same item (gh vs github, omitted vs concrete default host) no longer bumps the create generation, so an in-flight create's success lands its override instead of being discarded, and the button is not re-enabled for a duplicate request. Regression tests in both components rerender with `provider: "gh", platformHost: undefined` mid-create and assert the response still records the override (both verified to fail against the unfixed code).

Activity full-stack coverage: a new chromium e2e test in detail-action-buttons.spec.ts drives the real chain on the isolated workspace backend — open the activity feed, select the seeded PR row, create a workspace from the drawer detail, and assert the page stays on the activity selection URL while the workspace host and live terminal view attach inside Activity's own outer dock, then wait for the workspace to reach ready. This exercises the selection -> detail load -> claim (via the embedded list view's claim effect reacting to the recorded override) -> outer dock -> hosted terminal path that the component tests stub out.

Validation: full Vitest suite (3432 passed), svelte-check/lint/format clean, full chromium e2e (533 passed, 1 failure — the known stale-asset-recovery flake, passes in isolation).

wesm:
Round-13 finding addressed in 0664828f.

**Task-list edits fail locally for aliased providers or omitted default hosts** — fixed. The detail and issues stores now normalize body-edit identity comparisons through `canonicalProvider` and `resolvedPlatformHost` (a shared `sameBodyTarget` helper) in the local-body setters, unsaved-body guards, and body-PATCH response guards, and the save-queue keys are canonicalized so aliased and canonical addressing collapse to a single queue. Regression tests in `detail.svelte.test.ts` and a new `issues.svelte.test.ts` load a detail under canonical identity, apply a local body edit addressed via `gh`/omitted host, and assert the optimistic update and unsaved-body flag; both fail against the unfixed stores.

Validation: full Vitest run (3435 passed; one pre-existing load-flake in `PierreFileDiff.test.ts` that fails identically with the fix stashed and passes isolated), frontend-check clean, full chromium e2e (533 passed; known `mobile-routes.spec.ts:372` flake passed on isolated rerun).

wesm:
Round-15 finding addressed in 04d01636.

**Inline workspace terminals impossible to collapse during setup or error states** — fixed. The dock's collapse control lived only in the workspace toolbar, which renders once the workspace is ready, and WorkspaceDockPanel's BottomDock is intentionally not closable. `WorkspaceTerminalView` now renders a collapse button in the creating, fetch-failure, and setup-error branches whenever it is hosted in an inline dock; collapsing is pure local UI, so the control is never disabled by action guards. Four new tests in `WorkspaceTerminalViewEmbed.test.ts` cover collapse from each non-ready state plus the no-controller case; the three positive tests fail against the unfixed component.

Validation: full Vitest run clean when not sharing the machine with the e2e suite (unit 3306 passed, browser 161 passed — an earlier concurrent run showed load-contention failures across unrelated browser specs that all pass on rerun), frontend-check clean, full chromium e2e with only the known `mobile-routes.spec.ts:372` flake, which passed on isolated rerun.

wesm:
Round-16 finding addressed in 28962d46.

**Expanded mode disabled internal terminal pane resizing** — fixed. The expanded-mode rule hiding BottomDock's top-edge handle used a descendant selector, so it also hid the reparented terminal's nested pane handles. The hide is now scoped to the dock's direct child (`.kit-bottom-dock > .kit-split-resize-handle`). A new browser test in `WorkspaceDockPanel.browser.svelte.ts` pins that a handle nested in the dock slot keeps rendering while expanded while the dock's own handle stays hidden; it fails against the unscoped selector. The kit-ui re-audit guidance in `context/ui-design-system.md` now names the child combinator.

Validation: full Vitest run with only the pre-existing `PierreFileDiff` load flake (passes isolated), frontend-check clean, full chromium e2e with only the known `mobile-routes.spec.ts:372` flake (passes isolated).

wesm:
Round-17 finding addressed in 0d57325e.

**Failed inline workspace switch left a stale, uncollapsible terminal** — fixed, with all three prescribed changes:

- The toolbar's dock mode controls no longer sit behind `actionsBlocked`: mode changes are pure local UI and stay available during transitions and deletes; only the modal-stack guard remains, and only on the expand direction.
- The template now renders by liveness: the loading branch requires `workspaceLive` (payload matches the current id and host key), so an in-place A→B switch shows the loading state — which carries its own always-enabled collapse control — instead of A's cached ready toolbar, and a failed switch shows the error state (`loadError && !workspaceLive`) with retry and collapse.
- Two new tests in `WorkspaceTerminalViewEmbed.test.ts` cover a failed and a slow in-place switch (both fail against the unfixed component). One unit test and one e2e assertion that encoded the old stale-render contract were updated: they now expect the loading state while the new payload is gated and the details-loading sub-state only once it is live.

Validation: full Vitest 3470 passed (clean, including the usual PierreFileDiff flake this time), frontend-check clean, full chromium e2e rerun after the spec update with only the known `mobile-routes.spec.ts:372` flake, which passed isolated.

wesm:
Round-18 finding addressed in c9513fee.

**Duplicate workspace creation via remount/selection reset** — fixed. The Create Workspace button was gated on a component-local creating flag that the route-reset effect clears on identity change and that dies on remount, so returning to the item mid-request re-enabled the button. Pending creates now live in a module-scoped store keyed by canonical item identity (`packages/ui/src/stores/workspace-create-pending.svelte.ts`): `createWorkspace` refuses to start while that identity has a create in flight, the button disables and labels "Creating..." from the shared state, and the entry always clears in the request's `finally` so a settled create re-enables the action everywhere at once.

Regression tests: selection round-trip tests in both `PullDetail.test.ts` and `IssueDetail.test.ts` (click create, A→B→A before resolving, assert the button stays disabled, a second click sends no POST, and the eventual response records exactly one override), plus a remount variant in `PullDetail.test.ts`. All three fail against the unfixed components.

Validation: full Vitest 3473 passed with zero failures, frontend-check clean, full chromium e2e with only the known `mobile-routes.spec.ts:372` flake, which passed isolated.

wesm:
Round-19 finding addressed in 4ec77874.

**Confirmed creations published only through the optional inline controller** — fixed. `workspace-create-pending.svelte.ts` now also records confirmed creations by canonical identity, and controller-less detail instances (focus/mobile views, DetailDrawer) fall back to that record when their envelope carries no workspace, so a response landing after unmount or a selection round-trip flips the button to "Open Workspace" instead of re-offering a duplicate create. Records reconcile away when an identity-matched envelope carries the workspace (a null envelope is a stale pre-create fetch and keeps the record) and clear by workspace ID when `notifyWorkspaceDeleted` fires. Store mutations no-op without a state write when nothing matches — the reconcile runs inside the components' envelope effect, where an unconditional reassignment loops the effect.

Regression tests: no-controller selection round-trip tests in both `PullDetail.test.ts` and `IssueDetail.test.ts` (click create, A→B→A, resolve stale → "Open Workspace" renders, no Create button, no navigation, one POST); both fail against the unfixed components.

Validation: full Vitest 3475 passed (a one-off load flake in `App.test.ts` on the first run passed isolated and did not recur), frontend-check clean, full chromium e2e with only the known `mobile-routes.spec.ts:372` flake, which passed isolated.

wesm:
Round-20 finding addressed in 9e79bb7d.

**Inline views ignored the shared created-record** — fixed at the source. The host store's `effectiveRef` now falls back to `createdWorkspaceRef` after overrides and the envelope, so every consumer of `effectiveWorkspaceRef` — the detail buttons and the list views' claim effects that drive the dock — sees a creation that started in a controller-less focus/mobile view even when the layout switched to desktop before the response landed. Tombstones still win, and deletions clear the record by workspace ID, so a deleted workspace cannot resurface through the fallback.

Coverage: a new workspace-host store test records a creation with no controller involvement and asserts `effectiveWorkspaceRef` surfaces it (fails against the unfixed store), envelope stays authoritative when present, and identity scoping holds; a second test pins that a deletion still masks the record. The requester-side publish path is covered by round 19's no-controller tests.

Validation: full Vitest 3477 passed, frontend-check clean, full chromium e2e passed outright (535 passed, no flakes this run).

wesm:
CI run 29962576709 triage:

- **e2e (chromium) "Build frontend" failure is a flake.** The step died with `[lightningcss minify] Invalid empty selector` after vite-plugin-svelte failed to load a virtual CSS module — the identical build step succeeded in the `build` and `e2e (firefox)` jobs of the same run at the same commit, and the local production build passes. Nondeterministic; safe to re-run.
- **e2e mock (chromium/firefox) failures were real.** Two specs in `frontend/tests/e2e/workspace-sidebar.spec.ts` still asserted the pre-liveness stale-render contract (previous workspace's header/diff sidebar visible while the next detail load is held). 1db274f9 aligns them with the liveness rendering shipped in 28962d4's fix: the held window now asserts the "Setting up workspace..." loading state, and the cross-host spec keeps its guarantee that no member-host request carries the local workspace's id. Full mock lane (200 specs), full Vitest (3477 tests), and frontend-check pass locally.

mariusvniekerk:
ran some computer use simulation it and manually verified.

> Verified successfully with the e2e fixture database and Chrome Computer Use.
> - Issue #10 creates a ready inline workspace while remaining on the Issue page.
> - Collapse and “Focus Terminal” correctly restore the split dock.
> - A typed terminal marker survived “Open in Workspaces,” confirming session continuity.
> - PR #1 creates and hosts its workspace inline while remaining on the PR page.
> -  Activity selecting PR #1 displays its ready workspace inline.
> - Both worktrees use the expected branches.
> - Backend and frontend return HTTP 200.
> - Worktree remains clean; no source changes made.

wesm:
ok! I'll run a few more roborev turns and then merge

wesm:
Round 21 fixes in 639d7d4f:

- **Stale deleted workspace state**: added a shared monotonic lifecycle tick (`packages/ui/src/stores/workspace-create-pending.svelte.ts`) that orders creation confirmations against detail-envelope request starts. Both detail stores stamp every envelope-producing request at request start; `reconcileWorkspaceCreated` and the host store's override reconcile now clear on a null envelope when its request began after the confirmation — authoritative absence from a cross-client delete — while stale pre-create nulls still cannot wipe a confirmed creation. A 404 on the workspace fetch itself now reports through `onWorkspaceDeleted`, clearing tombstones, created records, and route memory. Regression tests cover both stores, both detail components, and the 404/transient-error split; all were verified to fail against the unfixed code.
- **Continuity test durability**: the inline-workspace continuity specs no longer treat a screenshot-hash change as proof commands executed. Each typed command now touches a marker file in the workspace worktree and the test asserts the file appears on disk — after the initial render and after each of the two reparents (a third command was added after the flip back). The `data-continuity` DOM tag remains the reparent-vs-recreate proof.
- Fallout: three mocked-create e2e tests left the real server's detail envelope reporting no workspace after their fabricated 202, which the new semantics correctly treat as a deletion; they now augment the detail response with the workspace ref, matching the real server's synchronous row insert.

Validation: full Vitest (3483 passed), frontend-check, full e2e-full chromium (534 passed; mobile-routes.spec.ts:372 failed in-suite and passed isolated — known flake), and the full mock lane (200 passed).

wesm:
Round 22 fixes in 3237b1d9:

- **404 after a successful load** (`WorkspaceTerminalView.svelte`): the 404 path now snapshots the identity from the cached envelope, passes it to `onWorkspaceDeleted`, and clears the cached workspace (stopping both polls) so liveness rendering shows the error state instead of the deleted workspace. Covered by a success-then-404 test driven through the SSE `workspace_status` listener, plus the existing transient-500 negative.
- **Tombstone vs fresh-ID recreation** (`workspace-host.svelte.ts::effectiveRef`): the tombstone branch now consults the shared created record — a record under a different ID supersedes the tombstone while the deleted ID stays masked, so a controller-less recreation surfaces after a layout switch.
- **Different-ID replacement envelopes** (`workspace-host.svelte.ts::reconcile`): a positive override now reconciles on any envelope whose request started after the override — same-ID confirmation, authoritative absence, or a replacement workspace under a new ID — so the dock can no longer stay pinned to a deleted workspace. Pre-override different-ID envelopes remain masked.
- **Atomic payload+tick** (both detail stores): replaced the max-retained tick with `applyEnvelopeAt`, which rejects any response whose request started before the currently applied envelope and otherwise assigns payload and tick together, across all six pull-detail and four issue-detail envelope paths. Interleaving tests: an older-started `/sync` response after a newer refresh (pulls) and two overlapping polling refreshes resolving out of order (issues).

All six new tests were verified to fail against the round-21 code. Validation: full Vitest (3488 passed), frontend-check, full e2e-full chromium (534 passed; mobile-routes.spec.ts:372 known flake, passed isolated), full mock lane (200 passed).

wesm:
That's a flake. Merging

wesm:
The firefox failure is real so fixing

wesm:
e6929f01 fixes the persistent `e2e (firefox)` failure. It was not a flake: `openTerminalPanel` in `00-inline-workspace-continuity.spec.ts` asserted an xterm canvas, but xterm.js only creates a canvas when its WebGL addon activates. Headless Firefox has no WebGL, so xterm silently falls back to its DOM renderer (no canvas element), making the assertion impossible — every firefox job failed deterministically on this test across all commits since the spec landed. The readiness wait is now renderer-agnostic (`canvas, .xterm-screen`); the marker-file and `data-continuity` assertions still carry the actual liveness proof. The ghostty test keeps its explicit canvas count (ghostty-web is canvas-based and initializes under headless Firefox — all three continuity tests now pass locally on firefox, which the ghostty test had never reached before due to serial-mode abort).

Validation: continuity spec 3/3 on firefox; full Vitest (3488 passed); full chromium e2e-full (534 passed, 1 failed: the known `mobile-routes.spec.ts:372` in-suite flake, passes isolated); frontend-check clean.

wesm:
c87c8bfe addresses the round-23 finding (delayed create response resurrecting a deleted workspace).

Verified the race: all deletions flow through `notifyWorkspaceDeleted`, but both creation publications were unconditional — `recordCreated` overwrote a same-ID tombstone with a positive override, and `recordWorkspaceCreated` republished the record `clearCreatedWorkspaceById` had just cleared. Fix per the finding: the shared create-pending store now persists deleted workspace IDs for the session (`markWorkspaceIdDeleted`/`isWorkspaceIdDeleted`), and both publication paths refuse a deleted ID. `notifyWorkspaceDeleted` and controller `recordDeleted` mark the ID; fresh-ID recreations are unaffected and supersede the tombstone as before (covered by a guard test that passes pre-fix by design).

Regression tests (workspace-host.test.ts): delayed create response cannot overwrite its own tombstone; cannot republish a deleted controller-less record (deletion with no identity/claim, where only the ID memory can block it); fresh-ID create supersedes the tombstone; `recordDeleted` blocks same-ID publications in both stores. 3 of 4 fail against the unfixed stores (fixes stashed), all pass with them restored.

Validation: full Vitest 3492 passed; frontend-check clean; full chromium e2e-full 533 passed — 2 in-suite failures (`mobile-routes.spec.ts:372` known flake, `activity-drawer.spec.ts:628`) both pass isolated; full mock lane 200 passed. One local e2e run had to be repeated after the known nondeterministic vite-plugin-svelte build flake left a stub dist.

wesm:
e5f882c5 addresses both round-24 findings.

Finding 1 (`reconcileWorkspaceCreated`): confirmed — the guard only covered the null-envelope case, so any non-null envelope cleared the record regardless of ID or tick; the host store's `CreatedOverride` got the full guard in round 22 but the shared-store twin did not. It now reconciles under the mirrored rule: same-ID envelope (server confirms the creation) or a request started after the confirmation (authoritative whatever it carries); a stale different-ID envelope no longer erases a recreation's record.

Finding 2 (controller-less resolution): confirmed — both components resolved `envelope ?? createdWorkspaceRef` with no deleted-ID filter. They now share `resolveControllerlessWorkspaceRef`: the confirmed created record wins until reconciled, an envelope still carrying a session-deleted ID is masked (those views subscribe to no invalidation, so the cached envelope outlives a deletion until the next fetch), and a fresh-ID envelope stays authoritative. `deletedIds` became reactive `$state` so a deletion updates a mounted view's button immediately rather than on the next unrelated rerender.

Six regression tests: the reconcile guard matrix (different-ID stale kept, with and without tick; same-ID clears regardless of tick; post-confirmation different-ID clears), resolver ordering/masking at the store level, and per-component controller-less tests for the deleted-envelope mask (live, via the reactive set) and created-record-over-stale-envelope preference. All six fail with the fixes neutered to the legacy behavior (exports kept so the failures are assertions, not import errors) and pass with them restored.

Validation: full Vitest 3498 passed; frontend-check clean; full chromium e2e-full 535 passed (no flakes this run); full mock lane 200 passed.

wesm:
3a91c53f addresses the round-25 finding.

Confirmed in both branches of `effectiveRef`, not just the tombstone one: the tombstone branch returned any different-ID envelope before consulting the shared created record, and the no-override fallback was `envelope ?? createdWorkspaceRef`. Either way a stale pre-confirmation envelope could shadow a controller-less recreation after a layout switch and let the dock claim a wrong or deleted workspace — including an earlier-generation deleted ID that no current tombstone remembers (tombstones hold only the latest deletion per identity).

Fix per the suggestion: both branches now resolve under the shared rule — the created record wins over a different-ID envelope until reconciliation (driven by the newer lifecycle tick) removes it; a same-ID envelope wins for its fresher status; and session-deleted envelope IDs are masked via the session deleted-ID memory, which covers earlier generations the tombstone cannot. The no-override branch delegates to `resolveControllerlessWorkspaceRef`, which gained the same-ID-envelope status preference (this preserves the existing "envelope carrying the workspace stays authoritative" contract).

New coverage: the requested controller-less-recreation-plus-inline-switch scenario with a stale earlier-generation envelope (two deletion generations, so the stale ID differs from the live tombstone); created-record-over-clean-different-ID-envelope until reconciled; the no-tombstone variant; and same-ID status freshness at both levels. Three existing tests codified pre-deletedIds behavior (a session-deleted ID resurfacing through an envelope after tombstone reconciliation, and a deleted ID surfacing under a different identity) and were updated to the strengthened semantics — deleted IDs never resurface, for any identity. Fail-first: 6 tests fail with the two store files stashed, all 131 pass restored.

Validation: full Vitest 3501 passed; frontend-check clean; full chromium e2e-full 534 passed (single failure was the known `mobile-routes.spec.ts:372` in-suite flake, passes isolated); full mock lane 200 passed.

wesm:
going for another round

wesm:
6f2a4f16 addresses the round-26 finding.

Confirmed: a release-and-reclaim within one update fell through all three existing reset paths — WorkspaceDockPanel's reset-on-inactive effect (effects only see the final claimed state, so no inactive gap is ever observable), its teardown reset (the panel never unmounts), and setClaim's replacement reset (no previous claim exists by then). The newly claimed item's detail opened hidden behind the fullscreen terminal.

Fix: `clearClaim` resets `expanded` to `split` synchronously, per the first suggested option. Focus restoration is preserved — the panel reclaims focus on the `dockOpen` open→closed transition, which any observed release still produces regardless of mode (verified before choosing this option; the reset-on-inactive branch it bypasses only handled the mode write plus a focus reclaim that the dockOpen effect also covers). The inactive and teardown resets remain as backstops.

Regression test: claim → expand → release + reclaim of a different identity in the same update → mode is split. Fails with the store fix stashed (stays expanded), passes with it restored.

Validation: full Vitest 3502 passed; frontend-check clean; full chromium e2e-full 534 passed (single failure was the known `mobile-routes.spec.ts:372` in-suite flake, passes isolated); full mock lane 200 passed.

wesm:
Queued merge pending CI

PR #744, opened 2026-07-25
fix: recover Roborev reviews after daemon restart

mjacobs:
Make the Reviews view recover promptly after the Roborev daemon becomes available again and keep its event stream disconnected during known outages. Recovery coverage now owns its daemon stop/start lifecycle.


wesm:
removing the superpowers docs

mariusvniekerk (approved):

PR #745, opened 2026-07-25
Remove Board mode

wesm:
Middleman no longer includes a separate Board surface for local PR workflow state. Board navigation, routing, shortcuts, and visible-mode settings are removed while workflow status remains available in PR detail, filters, API and CLI access, and local persistence.

- Keep PR workflow status editable and filterable
- Remove the dedicated Board and drag-and-drop UI
- Generate workflow screenshots during documentation builds instead of tracking generated SVGs
- Refresh current workflow documentation

## Screenshot

PR detail with workflow status and no Board navigation


wesm:
PR detail with workflow status and no Board navigation

mariusvniekerk (approved):

PR #746, opened 2026-07-25
Schedule GitHub sync from credential quotas

mariusvniekerk:
GitHub App installations and the user credential draw from independent REST and GraphQL pools. Treating them as one host-wide value could pause all background work while provider capacity remained available.

- Schedule GitHub background and archive work from the matching credential and API resource quota
- Keep `sync_budget_per_hour` as a separate 50,000-request local emergency ceiling
- Reconcile provider quota at most once every three minutes and distinguish provider quota from the local ceiling in status surfaces

Validation: repository hooks passed, including short Go tests, frontend checks, lint, generated API validation, and an isolated Chromium UI capture.


mariusvniekerk:
Addressed the two Medium findings on `7825241` in `61be64d34`.

**`sync.go:4256` (snapshot refresh cadence)** — fixed. The three-minute window is now claimed before the attempt rather than recorded after success, so a credential whose `/rate_limit` call keeps failing costs one request per window instead of one per pass, and a failure leaves the last known pool in the registry rather than blanking it. The fallback to a healthy route stays, because routes sharing an identity can hold distinct clients and one expired token must not stop the credential from refreshing through another; only distinct clients are tried, since repeating the identical client once per owner cannot produce a different answer. Covered by `TestRefreshRateLimitSnapshotsSpendsOneRequestPerCredentialWhenRefreshFails`, verified to fail against the previous mark-on-success code.

**`sync.go:3735` (reserve checked once at cost one)** — fixed for GraphQL, declined for REST.

The GraphQL half was real and is fixed in `a2a33435`: bulk pagination is unbounded and had no re-check at all, so a single large repository could page straight through the buffer. `GraphQLFetcher.backgroundReserveGuard` now re-checks between pages and aborts at the reserve. That abort is safe rather than lossy because any bulk-fetch error falls back to the REST index and only a successful fetch persists.

The REST half is an accepted soft-admission boundary, not a defect:

- Admission is re-evaluated per repository inside the worker loop (`sync.go:4475`), not once per pass, and REST responses update the pool as they arrive — so the overshoot is bounded by the repositories already in flight when the pool crosses, at most `defaultParallelism` (4).
- The hard guard is already per wire attempt: the local hourly ceiling is enforced atomically in `budgetTransport` (`TrySpend(1)`) before any request goes out. The provider reserve is deliberately the soft one.
- Enforcing it per REST attempt would convert a bounded overshoot into mid-sync refusals with partial persistence, which is worse than the overshoot and inconsistent with the detail drain's already-documented soft-admission model in `context/github-sync-invariants.md`.
- This granularity is inherited from the pre-existing host eligibility design; the change here swaps the host-wide pool for the routed credential's own pool, which strictly tightens isolation rather than loosening it.

Both invariants are recorded in `context/github-sync-invariants.md`.

mariusvniekerk:
All three Medium findings on `61be64d` are fixed in `01676f9ff`, plus the two from the local review on the same head.

- **Issue pagination bypassed the guard** (`graphql.go:1068`) — correct. `backgroundReserveGuard` now runs between issue pages too, and `FetchRepoIssues` suppresses the smaller-page retry for the reserve sentinel the same way `FetchRepoPRs` does. Covered by `TestGraphQLFetcherStopsBackgroundIssuePaginationAtGraphQLReserve`.
- **Unobserved in-memory quota bypassed persisted state** (`sync.go:3730`) — correct, and the sharpest of the three. Admission now falls back to the rate tracker's SQLite-backed pool when the registry has never observed a credential, so a restart no longer admits background work against a reserve the persisted state says is spent. An elapsed persisted reset window is ignored, since it cannot speak for the current quota. The same fallback applies to GraphQL admission, which returns to consulting the fetcher's tracker when the registry is silent. Covered by `TestBackgroundAdmissionUsesPersistedTrackerWhenRegistryIsEmpty` and `TestBackgroundAdmissionIgnoresPersistedTrackerPastItsResetWindow`.
- **Exhausted credentials monopolized workers** (`sync.go:4475`) — fixed, though narrower than described: repositories on an already-exhausted bucket are filtered before dispatch, so only a credential crossing its reserve *mid-pass* reached the in-worker gate. That gate now reports `skipped: rate limit throttled` and moves on, matching what dispatch already does, instead of holding a worker until reset. Covered by `TestRunOnceSkipsMidPassExhaustionWithoutHoldingAWorker`. Note the pre-existing tracker-backoff sleep below it is unchanged — that one is a provider `Retry-After`, not a local reserve trip, and predates this branch.

From the local review on the same head:

- **Client-identity dedupe defeated by production route construction** — correct, and a real defect in my previous commit. Startup builds one client per route, so pointer dedupe never collapsed routes sharing a credential. Routes now carry the token source's canonical string and the refresh walk dedupes on that. The scope-derived source key would not do: it names where a route was configured, not which credential answers for it. Covered at the layer the finding named, by `TestProviderStartupGivesRoutesSharingOnePATOneCredentialKey` in `cmd/middleman`, which drives real `buildProviderStartup` route construction and asserts owner routes sharing one PAT agree on the key while an independent credential does not.
- **Full-stack e2e for the same fix** — declined. The gap that finding correctly identified was that the test did not model production route construction; that is now covered where the wiring actually happens. Triggering the same two links through the API would re-prove them without adding distinct confidence, which is the tradeoff `CLAUDE.md` describes for duplicate full-stack coverage.

Every new test above was verified to fail against the code it guards.

mariusvniekerk:
Fixed in `7f8ebd803` — two of the three findings on `01676f9`, plus four from the local review on the same head.

- **Concurrent responses could raise `Remaining`** (`quota.go:66`) — correct. Requests run four-wide and finish out of order, so a header issued earlier could land after a later one and restore quota already spent. `ObserveHeaders` now only lowers a pool within one reset window; a new window replaces it outright, and `/rate_limit` snapshots still reconcile. Covered by `TestQuotaRegistryHeadersNeverRaiseRemainingWithinAWindow`.
- **Write snapshots deduped on the read credential** (`provider_startup.go:645`) — correct, and precisely stated. Mutations skip App candidates, so `App A → SHARED_PAT` and `App B → SHARED_PAT` write as one credential while their read chains differ, which lost the write bucket its dedupe. Routes now carry a separate `WriteCredentialKey` built from the non-App candidates. Covered by `TestProviderStartupSeparatesWriteCredentialKeyFromReadChain`.
- **REST reserve granularity** (`sync.go:4536`) — same finding as the previous head; the evidence-backed rejection stands, and one detail is now stronger than when I first wrote it. Admission is re-evaluated per repository inside the worker loop, the hard guard is the local ceiling enforced per wire attempt in `budgetTransport`, and per-attempt enforcement of the provider reserve would convert a bounded overshoot into mid-sync refusals with partial persistence. Since that reply, the *unbounded* half of this concern — GraphQL pagination, which had no re-check at all — has been closed on every paginated path including retries.

From the local review on `01676f9`:

- **Retry restarted past the guard** — correct. Every GraphQL retry now re-checks, including the native-stack-preview fallback. Covered by `TestGraphQLFetcherDoesNotRetryPastTheReserveWithASmallerPage`.
- **GraphQL restart fallback used `ShouldBackoff`, which only blocks at zero** — correct. It now applies the context-appropriate reserve and cost against the fetcher's persisted tracker, ignoring an elapsed window, exactly as REST admission does. Covered by `TestBulkGraphQLPersistedFallbackAppliesTheBackgroundReserve` and `...IgnoresAnElapsedWindow`.
- **Mid-pass skip left the bucket eligible for the drains** — correct, and a hole the sleep-to-skip change opened: previously the worker slept until reset, so the drains ran against a recovered credential. Workers now record the exhausted bucket and the drains see it as ineligible. Covered by an extension to `TestRunOnceSkipsMidPassExhaustionWithoutHoldingAWorker`, verified to fail without the propagation.
- **Full-stack API tests for the three reserve fixes** — declined. These are internal admission mechanics with no API surface of their own; the wire path they do have (`/api/v1/rate-limits`) is already covered against a real server and SQLite. Driving them through the API would re-prove the same links, which is the duplicate-coverage tradeoff `CLAUDE.md` describes.

Every new test was verified to fail against the code it guards.

mariusvniekerk:
Fixed in `baf5af14d` — both Medium findings from the local review on `7f8ebd8`, which sharpened two things I had gotten only half right.

- **Any different reset timestamp replaced the pool.** My monotonic rule treated a *different* window as a *newer* one, so a response from the window that had already closed could arrive last and rewind `ResetAt` to it. That matters more than it sounds: an elapsed window reads as expired, expired reads as unobserved, and unobserved admits background work — so a stale straggler could undo an observation that had just reached the reserve. Observations are now ordered by reset: older dropped, equal may only lower the count, later replaces the pool.
- **Worker exhaustion markers only existed when a later repository tried the credential.** The last repository on a credential can spend its headroom with nothing behind it, so no marker was recorded and the drains went on spending. Eligibility is now re-read after the workers finish and again between the detail and comment drains, folding the markers in rather than relying on them. The regression test for this is a single-repository case, since the earlier two-repository shape was what made a marker exist at all.

The recurring **full-stack HTTP-and-SQLite e2e** request is still declined, on the merits rather than as an omission. Production route construction — the gap earlier reviews correctly identified — is covered where the wiring happens, by two tests driving real `buildProviderStartup`, including the shared-PAT write dedupe. The wire surface these fixes have, `/api/v1/rate-limits`, is covered against a real server and SQLite. What is left is internal admission mechanics with no API surface of their own, and an API-triggered test would re-prove links that are each covered directly.

Both new tests were verified to fail against the code they guard.

mariusvniekerk:
Fixed in `9beb324c7` — the two Medium findings from the local review on `baf5af1`, with one half of the first declined on the merits.

- **`UpdateSnapshot` bypassed the ordering rule.** Correct, and fixed: a `/rate_limit` response issued before a burst of ordinary requests could land after them and rewind `ResetAt` into a window that had already closed — which reads as expired, which reads as unobserved, which admits work the current window had ruled out. Older-window snapshots are now dropped.

  I declined the other half of the suggested fix — making equal-window snapshots lower-only, as response headers now are. A snapshot is not a header: refreshes are claimed so only one is in flight per credential, and it is the endpoint whose purpose is to state the current truth. Lower-only would leave a pool that out-of-order headers clamped too low stuck there until the window rolls, over-throttling for up to an hour with no way to correct it. The pre-existing `TestQuotaRegistryReserveRequiresEveryKnownPool` also encodes equal-window snapshots as authoritative in both directions. The harm the finding identified was the `ResetAt` rewind, and that is closed.
- **Drains rechecked eligibility only once at the start.** Correct. Both loops now re-read the reserve per item, immediately beside the local ceiling check each already performed, and revoke the bucket so the rest of its items stop too. Covered by `TestDetailDrainStopsAtTheReserveItCrossesPartway`, verified to fail without it.

The full-stack e2e request is declined again on the same recorded evidence.

mariusvniekerk:
Two more heads since the last note.

`1b4b0c421` — the comment drain's per-item reserve check covered only its pull-request loop; the issue loop spends the same credential, so an issue-only queue kept refreshing past the reserve. Both loops now check, and comment refresh is admitted at a cost reflecting that a thread is fetched through paginated calls rather than one request.

`a04592750` — the Medium finding on `9beb324` (`notifications_sync.go:461`) was correct: notification admission read only the in-memory registry, so after a restart a three-request acknowledgement could start on 201 remaining and cross the reserve, exactly when a failed `/rate_limit` refresh keeps the registry silent. It now falls back to the persisted write tracker, mirroring what repository admission already did. Covered by `TestNotificationBudgetUsesPersistedWriteTrackerAfterRestart`, verified to fail without the fallback.

**Not addressed — the High finding on `9beb324`** (atomic per-pool in-flight reservations before every background wire attempt). This is the same reserve-granularity position I've declined three times with recorded evidence, now escalated in severity. My judgment is unchanged and I'd rather state it plainly than keep relitigating it:

- The reserve is a soft admission buffer, not a lease. The hard guard is the local hourly ceiling, enforced atomically per wire attempt in `budgetTransport`.
- The genuinely unbounded paths this finding pointed at have been closed one by one over the last several heads: GraphQL pagination and its retries, both drain loops per item, worker admission per repository, post-worker recompute. What remains is bounded by the work already in flight when a pool crosses.
- Per-attempt provider-reserve enforcement would convert a bounded overshoot into mid-sync refusals with partial persistence, which is worse than the overshoot.

One part of that finding I have **not** verified and am flagging rather than dismissing: the claim that background detail paths can issue GraphQL timeline requests while `backgroundQuotaAvailability` checks only REST. If that is accurate it is a real gap of the same kind as the ones already fixed, and it deserves its own look.

I've reached this run's iteration limit, so I'm stopping here rather than continuing to push.

PR #747, opened 2026-07-25
Remove the Messages mode and msgvault integration

wesm:
The Messages surface now lives outside middleman, so the in-tree mode is removed.

- Remove the Messages app mode: the `/messages` route, feature shell, components, saved views, and setup dialog
- Remove the msgvault backend adapter, messages API routes, saved searches, and remote image proxying
- Remove the `[msgvault]` config block and the `messages` mode-visibility toggle; existing configs with those keys still load, the keys are ignored
- Remove the kata linked-message UI (message links stored in Kata task metadata are left untouched)
- Regenerate the OpenAPI spec and generated clients
- Update user docs and agent context to drop Messages/msgvault

## Screenshot

App header with Kata and Docs tabs and no Messages tab


wesm:
Addressed both findings in 27f64fed.

**Finding 2 (regression coverage for removed surfaces):**
- `internal/server/messages_removed_test.go` asserts at the wire level (through `srv.ServeHTTP`) that the former API routes — `GET /api/v1/msgvault/health`, `POST /api/v1/msgvault/configure`, and `GET`/`PUT /api/v1/messages/saved-searches` — all return 404.
- `frontend/tests/e2e-full/navigation.spec.ts` adds a full-stack test against the live e2e backend (which runs with imported modes visible): `/messages` falls back to the Activity feed with no Messages shell, the top bar offers Kata/Docs tabs but no Messages tab, and the Settings "Visible modes" panel has no Messages toggle.

**Finding 1 (legacy `[msgvault]` config / `api_key_env` stripping):** maintainer decision is to ship no legacy handling at all — no rejection error and no decode-only sanitization. A leftover `[msgvault]` section is ignored like any other unknown config key, and operators should delete it along with the env var it referenced. Middleman never persisted the credential itself (only the env var name), so the removal does not leave secrets behind in middleman state.

Validation: `go test ./internal/config ./internal/server`, `golangci-lint` (0 issues), full Vitest unit (2955) and browser (170) suites, and the full Playwright e2e-full suite (459 passed; one unrelated terminal-font settings flake passed on rerun).

PR #748, opened 2026-07-25
Focus terminals on agent launch and add explicit create-and-launch choices

wesm:
## Why

A persistent default-agent setting plus a confirmation modal made “auto-launch” contradictory, and its PR-only trust gate differed from the existing manual Launch action. Workspace creation should remain neutral unless the user explicitly chooses an agent for that creation.

## What changed

- Focus newly created xterm and Ghostty terminal panes when an agent session launches, without stealing focus during later layout changes.
- Replace automatic launch with one shared **Create Workspace** split button across PR, issue, ad-hoc, and Kata creation surfaces.
- Keep the primary segment create-only; choosing an agent from the dropdown creates and launches it immediately with no confirmation modal.
- Carry the explicit choice as reactive, one-shot, in-memory intent keyed by workspace ID.
- Show unavailable agents disabled with their reason and support menu keyboard navigation, focus return, and narrow detail drawers.
- Remove the unreleased `default_agent` configuration/API contract and the `require_same_repo_head` launch precondition.
- Keep `mr_head_repo_kind` as informational classification and make MR reclassification identity-stable during repository reconciliation.

## Trust boundary

An explicit create-and-launch choice for a fork or unresolved-head PR is equivalent to choosing the agent from the manual Launch menu. It may execute contributor-controlled repository hooks; no agent launches automatically.

## Verification

- Focused frontend creation/launch tests and full frontend formatting, lint, kit, and Svelte checks passed.
- Affected workspace, GitHub, and DB packages plus `go vet ./...` passed after rebasing onto current `main`.
- Repository reconciliation regressions passed 20 shuffled repetitions.
- Whole-branch review completed with no Critical, Important, or Minor findings.


wesm:
Response to the roborev combined review (`caffc9d`), addressed in `c9852dc6` and `63d750ac`:

**Default agents auto-launch in untrusted PR worktrees — fixed.** The workspace API now exposes the existing `db.Workspace.MRHeadRepo` tri-state as `mr_head_repo_kind` (`same_repo` / `fork` / `unknown`), and the auto-launch effect skips pull-request workspaces unless the head is confirmed same-repo. The skip burns the one-shot marker, so an untrusted workspace can never auto-launch later either; launching an agent there requires the maintainer's explicit click on a launch card, which is unchanged from pre-branch behavior.

**Workspace creation and agent auto-launch lack full-stack coverage — declining, per repo testing policy.** The runtime launch path already has real-server coverage: `internal/server/` runs ~25 E2E tests against the in-process HTTP server with SQLite that launch agent and shell targets through `POST /workspaces/{id}/runtime/sessions` (including agent-context writes, unavailable targets, multi-launch/stop, and natural agent exit), and the settings round-trip for `default_agent` is covered in `settings_test.go`. The trigger itself is frontend-owned and covered by behavioral component tests including launch-exactly-once and the new fork-gating cases. CLAUDE.md's testing guidance explicitly says not to add duplicate full-stack e2e when a real-backend server test proves the runtime path and component tests prove the UI behavior.

**Automatic terminal focus can steal focus — fixed.** Focus no longer lives in the `active` effect at all; each pane focuses exactly once, at terminal creation, when it is active and enabled. Reveal-driven `hostVisible` flips and `disabled` transitions can no longer move focus, so the opt-in `pendingHostFocus` / `WorkspaceDockPanel` focus contracts are preserved. The browser regression test now asserts an outside control keeps focus when an already-initialized pane becomes active.

wesm:
Addressed all three findings from the `63d750a` review in a8bef530..0a45c379:

**Stale trust classification (High)** — a8bef530. `refreshWorkspaceHeadRepo` now persists the recomputed classification back to `middleman_workspaces.mr_head_repo` whenever it differs from the stored value, via a new `db.UpdateWorkspaceMRHeadRepo`. Persistence failure is fatal to the operation that triggered the refresh: setup/retry fails to `error` instead of reaching `ready`, and agent launch prep returns the error, so a stale-permissive `same_repo` row can no longer back the API field or the auto-launch gate. Regression test: `TestWorkspaceRetryLegacyUnknownHeadRepoLeavesBranchUntrackedE2E` now starts from exactly the stale shape (stored `NULL` vs. actual unknown head) and asserts both the persisted row and the wire response report `unknown` after retry (previously red with `expected: "unknown", actual: "same_repo"`).

**Auto-launch on reused workspaces (Medium)** — 0a45c379. The issue and Kata create endpoints are idempotent (202 with the pre-existing workspace from both the pre-check and the UNIQUE-race fallback), so the response now carries `created: true` only on the genuinely-new paths; reuse paths omit it. All three frontend call sites (`IssueDetail`, `PullDetail`, `KataWorkspace`) gate `markWorkspaceForAutoLaunch` on that field. Covered by first-vs-second-POST assertions in `TestCreateKataWorkspaceReusesExistingScopedTaskWorkspace` and `TestWorkspaceCreateIssueIsIdempotent`, plus positive/negative component tests per call site.

**Coverage (High)** — the launch-workflow/trust-boundary path is now exercised end to end through the real HTTP server and SQLite in the repo's sanctioned lane (`internal/server` E2E via `srv.ServeHTTP` and the generated client): `mr_head_repo_kind` is asserted on ready responses for `same_repo` (`TestWorkspaceCreateSameRepoHeadCloneURLTracksOriginBranchE2E`), `fork` (`TestWorkspaceCreateGitLabUsesSpecificMergeRequestHeadRefE2E`), and `unknown` (the retry regression test above), and the created-vs-reused discriminator is asserted over the wire in the two idempotency tests. The frontend one-time-launch gating (trusted-new launches once; fork/unknown, reused, and already-active do not) is covered by the component and browser tests on `WorkspaceTerminalView` and the three call sites against these real response shapes, per the repo's testing-lane policy of not duplicating a proven server path with full-stack browser tests.

wesm:
Addressed both findings from the `0a45c37` review in 77617506 and bcc00aab:

**Focus stolen after async init (Medium)** — 77617506. Both panes now capture focus intent at mount (`terminal-focus.ts`): `document.activeElement` is recorded before the async init window (xterm's font-load race, ghostty's WASM init), and the terminal takes focus at `start()` completion only when that element is still focused and is not a form control or inside an open `[role="dialog"]`/`[role="menu"]`/`[role="listbox"]`. This cancels the steal when the user focuses another control mid-init, and covers the live-preview renderer swap specifically: the options popover is `role="dialog"`, so a pane remounted under it never grabs focus. Plain buttons are intentionally not protected — clicking a launch tile focuses a button, and the newly created terminal must still take focus in that case (the two existing browser focus tests still pass). New coverage: unit tests for the guard plus browser tests for both panes asserting focus stays on an externally focused button and inside an open dialog across terminal init.

**Unavailable default agent (Medium)** — bcc00aab. Two changes: (1) the default-agent picker now consumes launch-target availability — the `launch_targets` field already present on `GET /settings` (computed by the same `localruntime.ResolveLaunchTargets` resolver as the workspace runtime endpoint, real `exec.LookPath` checks), so no new endpoint was needed; unavailable agents are annotated with the server's `disabled_reason` (e.g. "codex not found on PATH") and disabled in the dropdown, except that a currently-saved unavailable default stays selectable so it remains visible and replaceable. (2) When workspace readiness consumes the one-shot marker but the configured default is missing, not an agent, or unavailable, the view now shows a danger flash naming the agent and reason instead of returning silently — the deliberate no-ops (fork-PR gate, no default configured) stay silent. Covered by component tests for the picker annotation and for the flash-without-launch path.

wesm:
Addressed both findings from the `bcc00aa` review in bfd2e7af..9e44f1b5:

**Stale workspace trust / auto-launch bypass (High)** — fixed on both fronts the review named, with one deliberate scope refinement.

- *Sync-time synchronization* (bfd2e7af, 9e44f1b5): every MR snapshot upsert now fans out to a workspace trust reclassification. The hook lives in `commitMergeRequestParentSnapshot` — the single choke point all MR-upsert paths funnel through (REST list sync, GraphQL bulk fetch, detail refresh, closed-MR refresh; all 8 call sites traced) — and recomputes `mr_head_repo` from the post-upsert stored row, so a list sync reporting an unknown head cannot downgrade an already-resolved classification. Failures log rather than abort the sync, matching other non-fatal per-item persistence steps. Covered by a sync-path regression (`TestCommitMergeRequestParentSnapshotReclassifiesWorkspaceHeadRepo`, red without the fix) plus an E2E asserting a ready same-repo workspace reports `fork` after a sync retargets the MR.
- *Server-side precondition* (2c800bbc): the launch endpoint accepts `require_same_repo_head`; when set, it refreshes the classification via the newly exported `workspace.Manager.RefreshWorkspaceHeadRepo` immediately before deciding and rejects with 409/`untrusted_head_repo` for a PR workspace whose head is not confirmed same-repo — no session is created and the row is persisted with the corrected value. Only default-agent auto-launch sends the flag. A blanket launch gate would be wrong here: manually launching an agent in a fork workspace is a deliberate, allowed maintainer action (the round-1 decision gated auto-launch specifically), so manual launches are unchanged — and the E2E regression covers exactly the requested scenario: ready same-repo workspace, MR retargeted to a fork post-ready, auto-launch-style request refused (with the stored row reclassified) while the manual request still succeeds.

**Config hot reload misses default_agent (Medium)** — 1f72d6d4. The `config_changed` hydration in `Provider.svelte` was extracted into a testable `settings-hydration` module that now applies `default_agent` to the settings store, mirroring the startup path, with unit tests covering the new field and the previously hydrated ones.

PR #751, opened 2026-07-25
Route trusted Linux builds to managed runners

wesm:
## Summary

- route canonical same-repository Linux pull requests and main-branch builds to managed Linux runners, with forks and noncanonical copies falling back to GitHub-hosted runners
- invoke the reusable workflow from `main` and let checkout use the caller event's immutable SHA without accepting a dispatcher-supplied ref
- retain the high-concurrency managed test profile while using a conservative profile for hosted fork fallbacks
- preserve the caller permissions required to publish checks and packages; GitHub still downgrades fork pull-request tokens to read-only
- lint proposed workflow revisions on GitHub-hosted infrastructure

The managed runner group admits only the listed reusable workflow file from `main`; PR-controlled dispatcher changes therefore cannot target it directly. The reusable-workflow bootstrap is incorporated in this branch.


wesm:
false positive

PR #756, opened 2026-07-27
Show live agent state in workspace rows

mariusvniekerk:
## Summary

- receive Claude and Codex lifecycle hooks through a small middleman CLI subcommand
- distinguish working, idle, input, and approval states for live workspace sessions
- show quiet agent-state indicators in workspace sidebar rows while retaining tmux fallback

## Screenshot

Workspace sidebar showing Working, Input, and Approval agent states

<sup>generated by a clanker</sup>

PR #758, opened 2026-07-27
Refactor agent activity for maintainability

mariusvniekerk:
Pure refactor of the agent activity code that landed in #756. No behavior, API, or UI change.

- Installed hooks and the server resolve the report directory through one function instead of deriving the same path separately
- Single home for the hook source marker, the atomic file write, and the environment override merge, each of which had been duplicated
- Hook lifecycle events resolve through a table and a named outcome instead of a `(State, bool, bool)` return
- Workspace responses attach agent state through one explicit call instead of named returns with deferred mutation
- Agent-hook CLI moved out of `main.go` into `cmd/middleman/agent_hook_cli.go`, and the workspaceapi agent-activity helpers now live in one file

### DIFF
--- a/cmd/middleman/agent_hook_cli.go
+++ b/cmd/middleman/agent_hook_cli.go
@@ -1,0 +1,166 @@
+package main
+
+import (
+	"flag"
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+	"strings"
+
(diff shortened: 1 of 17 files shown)


mariusvniekerk:
bigger not smaller. can delete

PR #764, opened 2026-07-27
Report archive close and merge activity

wesm:
Archive reports currently project item creation but not closure and merge activity, so a completed archive cannot serve as a complete historical activity source. Existing archives may also predate provider-supplied lifecycle actors and merge metrics.

This adds a versioned lifecycle report contract, records issue and merge-request inventory coverage independently, and requeues canonical hydration once when archived records lack required lifecycle details. It also fixes two archive reliability issues found during end-to-end operation: direct CLI access behind trusted proxy configuration and bodyless 304 responses during durable inventory scans.

The repository commit and push hooks pass, including short tests, generated API validation, lint, and nil analysis.


wesm:
Addressed in d1f6775c:

- The daemon-aware CLI now round-trips the report schema, lifecycle kinds and totals, inventory coverage, actors, and merge metrics, with CLI contract coverage.
- Matching repository-feature-disabled inventory outcomes complete as unsupported before returned feature deferrals are honored; admission completion still runs.
- Lifecycle contributor counts use the close or merge actor.
- SQLite-to-HTTP generated-client coverage now asserts lifecycle actors, metrics, totals, and inventory coverage.

The full local repository gate and push hooks pass.

wesm:
Addressed in 3aba0225:

- Successful canonical lookup now advances a durable lifecycle-detail generation even when the provider legitimately omits an actor or merge metric. Legacy completed rows below that generation requeue once; pending and failed work keeps its normal retry behavior.
- GitLab lifecycle backfill is enabled alongside GitHub. Detailed merge requests preserve `merge_commit_sha`, with `squash_commit_sha` as the fallback.
- Regression coverage proves unavailable fields do not trigger a second backfill and exercises GitLab normalization through SQLite archive reporting.

The full local gate and push hooks pass.

wesm:
Addressed in 41bdffc4.

Archive issue hydration now bypasses the persisted parent-detail ETag, so a cached `304` cannot suppress the timeline refresh and then mark lifecycle data complete. Ordinary issue detail refreshes retain the conditional ETag path.

The regression seeds a legacy closed issue with a cached ETag and no lifecycle event, makes the conditional client return not-modified, and verifies that archive hydration performs the unconditional parent fetch, refreshes the timeline, and persists the close actor. The full local short suite, lint, vet, nilaway, context check, and commit/push hooks pass.

wesm:
Addressed in ef6c283f.

Capability reconciliation now treats persisted `unsupported` inventory as reopenable when the current provider advertises that stream. It atomically advances the scan to a fresh generation, clears cursors and errors, returns coverage to `unknown`, and lets the ordinary initial inventory path establish supported coverage. No migration, compatibility path, or periodic probe was added.

The regression exercises the full repository-feature transition: disabled inventory completes as unsupported, reconciliation after re-enablement reopens it, and the next eligible run completes it as supported. The full local short suite, lint, vet, nilaway, context check, and commit/push hooks pass.

wesm:
Addressed the three findings from `ef6c283` in `d6bb2457`:

- Archive issue hydration now propagates timeline failures in archive-budget contexts, while ordinary issue refresh retains its best-effort behavior. Lifecycle completion therefore cannot advance after a transient timeline failure.
- Repository-feature-disabled responses are classified before maintenance deferrals. The current maintenance stream completes as exhausted without rewriting previously established historical coverage.
- Lifecycle backfill now requires the latest authored close event to match the issue current `closed_at`. Report attribution also fails closed when the newest authored event belongs to a different close cycle.

Regression coverage includes transient timeline failure, disabled prompt maintenance, earlier close/reopen cycles, and an inconsistent newer close event. The full local gate passes: 5,515 tests, lint, vet, nilaway, and the context structural check.

wesm:
Addressed the `d6bb245` review finding in `79b0b33c`. GitLab issue events now read the full discussion stream and retain authored assignment, close, and reopen system notes. The ordinary-comments path remains separate and excludes system notes. Regression coverage drives a closed issue from a GitLab HTTP fixture through canonical archive hydration and SQLite into a detailed report, where the current close actor is asserted.

The linked CI failure was separate from archive reporting: `TestCaptureTelemetryEvent_QueuesEvent` left a server lifecycle running while its SQLite temporary directory was removed. `ed747c96` registers graceful server shutdown after database creation so cleanup drains background workers before SQLite teardown.

The exact CI test command (`go test ./... -shuffle=on -parallel=4`) passes locally after both fixes. The telemetry handler group also passes 50 consecutive runs; lint, vet, nilaway, and context checks pass.

wesm:
Addressed in `87f96730`. Thin clients now always synthesize `X-Forwarded-Host` from the discovered daemon listen address and keep it scoped to that same origin; daemons started without proxy trust ignore the header, while trust-enabled daemons remain reachable if the config file changes before restart.

The archive CLI end-to-end test now starts a trust-enabled daemon, changes the on-disk setting to disabled, and verifies the thin-client request still succeeds against the startup-bound server policy. The full shuffled Go suite, lint, vet, NilAway, context validation, and commit/push hooks pass locally.

wesm:
Addressed in `e7346f48`. Gitea now maps the SDK counters into the neutral merge-request model. Forgejo’s API emits the same fields but its latest SDK type omits them, so the existing bounded pull-request response capture now preserves the counters with head/base revision binding before normalization; this adds no provider request.

A provider-to-SQLite-to-report integration test covers both providers and asserts additions, deletions, and changed-file counts on merged activity. The full shuffled Go suite, provider suites, lint, vet, NilAway, context validation, and commit/push hooks pass locally.

wesm:
Addressed both findings in `60620555`.

- Lifecycle-detail reconciliation now gates on persisted issue/merge-request archive coverage instead of provider names, so completed Gitea and Forgejo rows receive the same bounded one-time refresh when that inventory capability is supported. Regression coverage exercises completed GitLab, Gitea, and Forgejo rows.
- The daemon CLI now rejects missing or non-current archive report schemas before converting any fields. Tests cover missing, future, and current-schema responses.

Local verification passed: the full shuffled Go suite, lint, vet, NilAway, and the context structural check.

wesm:
Addressed both findings in `c91ee67f`.

- Known-item lifecycle repair no longer depends on historical inventory coverage, so GitLab merge-request rows can backfill lifecycle details even though historical MR enumeration is unsupported. When an unsupported inventory stream is re-enabled, capability reconciliation now advances completed known-item lookups into a fresh fenced generation immediately. Regression tests cover both cases.
- Additions and deletions now carry explicit provider-presence metadata through Gitea/Forgejo capture, neutral normalization, and persistence. Omitted counters preserve stored values; fresh values and explicit zero replace them. Regression coverage exercises changed and zero metrics across successive sync snapshots, plus raw provider presence handling.

Local verification passed: the full shuffled Go suite, lint, vet, NilAway, hook checks, and the context structural check.

PR #765, opened 2026-07-28
Restore tmux clipboard copy and remove ghostty-web

wesm:
- Restore keyboard and mouse tmux copy in inline and tab workspace terminals.
- Preserve edge-drag autoscroll while selecting terminal output.
- Support Firefox and Zen through the native local clipboard fallback when browser clipboard writes are denied.
- Use xterm exclusively for workspace terminals and remove the terminal renderer setting.


wesm:
Addressed by testing and documenting the actual tmux trust boundary instead of adding gesture gating, which would break keyboard-only and custom copy bindings.

Middleman connects the browser to a tmux client. With tmux default `set-clipboard external`, a pane application OSC 52 is rejected, while tmux clipboard commands still reach the browser. The new real-browser test proves both paths: pane-emitted OSC 52 leaves the clipboard unchanged, and a keyboard-issued `set-buffer -w` updates it. As a sensitivity check, changing only the isolated test server to `set-clipboard on` made the negative assertion fail with the untrusted value.

wesm:
The clipboard integrity finding is now addressed in `3072a1af`. This supersedes my earlier tmux-filter-only response.

Xterm opens one deferred clipboard write from a trusted pointer or keyboard gesture, and the next accepted OSC 52 consumes it. Writes without a live authorization are ignored, reads remain rejected, and the direct `writeText` fallback is reachable only inside that one-shot authorization. Tmux still uses `set-clipboard external` as a second boundary against pane-application OSC 52.

Firefox coverage now exercises a keyboard-triggered tmux copy whose OSC 52 arrives six seconds later, ordinary drag copy in both workspace hosts, and a seven-second drag past the terminal edge. Edge scrolling emits only clamped SGR wheel/drag reports during an observed tmux drag and synthesizes release only when xterm does not, allowing tmux to finalize the selection.

wesm:
Addressed the Ghostty compatibility finding in `49c5d701`: Ghostty now intercepts bounded, write-only OSC 52 before its renderer, uses the same one-shot clipboard writer as xterm, and reaches the Firefox/Zen loopback fallback. The real Firefox/tmux test now selects `ghostty-web` and proves that path.

We are accepting the generic-interaction finding as an explicit threat-model tradeoff, without a code change. Middleman treats a process in a workspace terminal as native-terminal-equivalent: such a process can invoke `tmux set-buffer -w` just as a process in a native terminal can request an OSC 52 clipboard write. The browser-side gate prevents idle/background writes, bounds the payload, rejects reads, and permits only one write per recent terminal interaction; it is not intended to authenticate the semantic intent of an arbitrary tmux binding.

Restricting authorization to a dedicated UI command or known binding would break the required keyboard-only behavior and custom tmux key tables. `set-clipboard external` remains a separate boundary against raw pane-emitted OSC 52, while deliberately invoking tmux clipboard commands from the terminal process remains inside the accepted native-terminal-equivalent trust boundary.

wesm:
Addressed in `bc064fc9`:

- Removed clipboard authorization from Ghostty’s `term.onData` callback. That callback can carry terminal-generated responses, so clipboard authority now comes only from trusted DOM keyboard, pointer, paste, and composition events. The unit regression explicitly verifies terminal data cannot authorize a write.
- Made the Firefox coverage deterministic in headless CI by forcing browser clipboard denial and asserting the text posted to the CSRF-protected loopback fallback. The keyboard-copy, drag-copy, and edge-autoscroll paths now prove the fallback payload without requiring a Linux display or clipboard command.
- Isolated the tab and inline tmux drag checks by closing the first terminal WebSocket and reasserting mouse mode after the next host mounts; concurrent tmux clients otherwise affect sizing and mode delivery.

The Firefox e2e job in the linked run was already green. The two linked failed jobs were unrelated Go timeouts in `cmd/middleman TestAPIVerbE2E` and `internal/github/syncertest TestSyncerTriggerRunForReposSyncsOnlySelectedRepos`; this clipboard-only PR does not change those paths.

wesm:
Addressed in `2c912984`.

The finding reproduced for incomplete OSC command prefixes `5` and `52`: the parser forwarded the embedded introducer, allowing the nested OSC 52 sequence to reach Ghostty without passing through the gated handler. `osc-command` now preserves only the rejected outer prefix and restarts recognition at either a 7-bit `ESC ]` or C1 OSC introducer.

Regression coverage exercises both prefixes with both introducer encodings. The focused filter suite passes 9 tests, the complete frontend suite passes 3,287 tests with 2 skips, and the Chromium/Firefox clipboard integration passes 7 tests with its expected Chromium skip.

wesm:
Addressed in `ff9b86a4`:

- OSC parsing now treats C1 OSC immediately following `ESC` as a fresh candidate, so combined introducers cannot reach Ghostty unfiltered.
- Pointer clipboard authorization uses pointer capture and is revoked on lost capture, cancellation, window blur, document visibility loss, or a 60-second watchdog.
- Xterm records tmux mouse state only after input is delivered to an open WebSocket and resets drag/autoscroll state on connect and close.
- The Windows `clip.exe` fallback now sends UTF-16LE, with exact non-ASCII and surrogate-pair coverage.

Verification: full frontend suite passed (3,296 tests; 2 expected skips), the affected Chromium/Firefox Playwright suite passed (7 tests; 1 expected skip), `make frontend-check` passed, the system clipboard Go tests passed, and the package cross-compiled for Windows/amd64.

wesm:
Addressed in `d939ed0c`:

- Filtered OSC 52 sequences are now replaced with CAN as soon as the command is recognized. This prevents preserved `5`, `52`, and `ESC` prefixes from joining attacker-controlled suffixes, including when the inner sequence is incomplete and the filter resets across reconnect.
- Pointer-origin authorization remains tracked through the one-second release grace, so blur, visibility loss, cancellation, or capture loss revokes it after `endPointerGesture()` as well as during an active drag.

The regressions were verified red before the fixes. Final verification: 3,301 frontend tests passed, `make frontend-check` passed with zero findings, and the Chromium/Firefox tmux Playwright suite passed 7 tests with 1 expected skip.

wesm:
Addressed in `7a887dea`:

- OSC identifiers are parsed as bounded numeric commands, so leading-zero `052` forms are filtered for both 7-bit and C1 introducers; pathological overlong identifiers are canceled and discarded.
- Keyboard authorization created during an active pointer gesture now has its own deadline and remains pointer-cancelable, including after the original pointer authorization was consumed.
- The host clipboard route now rejects requests whenever `trust_reverse_proxy` is active, because a loopback proxy connection cannot prove that the originating browser is local.

Verification: 3,304 frontend tests passed; frontend checks passed; the Chromium/Firefox tmux suite passed 7 tests with 1 expected skip; focused clipboard server tests passed. The first broad server/short-test runs hit unrelated temp-directory/PTY lifecycle flakes; the commit-hook rerun passed the complete short suite.

wesm:
Addressed in `be4ddcde`:

- Before retaining a nested ESC or C1 OSC introducer, the filter now emits CAN after any forwarded outer prefix. A reconnect/reset can therefore discard retained bytes without leaving Ghostty in partial OSC state. Regression coverage exercises reset after `ESC ] 5 ESC` and after the equivalent C1 introducer.
- CAN and SUB inside buffered OSC 52 now abort the sequence immediately, discard its clipboard data, and return to normal parsing so following visible output is not swallowed.

Verification: 3,308 frontend tests passed; `make frontend-check` passed with zero findings; the Chromium/Firefox tmux Playwright suite passed 7 tests with 1 expected skip.

wesm:
Declined. rationale incoming

PR #771, opened 2026-07-28
Prevent archive catch-up from exhausting GitHub quota

wesm:
## Why

`sync_budget_per_hour` is a process-local emergency guard, not a statement of provider capacity. Archive catch-up previously scaled its quadratic allowance from that local number, so an operator setting a high guard could let historical hydration consume nearly all of a GitHub user credential’s 5,000-request quota. The provider reserve alone was insufficient because normal sync could consume the small amount left behind.

Archive work needs both a provider-aware pacing envelope and a hard bound from the credential’s current headroom. Without the latter, one already-admitted hydration could continue sending requests after a required pool approached its reserve.

## Approach

Archive admission derives a conservative window from every required routed-credential resource, using the smallest quota limit, smallest current remaining value, and latest reset. Its allowance is capped by the local surplus, the paced provider surplus, and `Remaining - reserve`. Unknown or expired resource windows defer archive work instead of falling back to the oversized local guard.

## Validation

- Unit coverage verifies the provider limit caps a 100,000 local guard and current remaining quota preserves the 200-request reserve.
- An API-to-SQLite server e2e starts archive work through `/archive/start`, runs a real quota-accounted GitHub transport against a fake upstream, and verifies ten attempted calls stop after three when `Remaining=203`. It also verifies the quota pool lands at 200 and SQLite records retryable interrupted hydration.
- The full GitHub and server e2e packages, repository short test suite, lint, commit hooks, and push hooks pass.
- The compatible branch build remains deployed on the affected laptop with a 3,000 local ceiling.


wesm:
Addressed both medium findings in `b9e37468`.

- `QuotaPacingWindow` now carries the smallest current remaining value across every required resource, and archive admission clamps each request allowance to `Remaining - provider reserve`.
- Added an API-to-SQLite server e2e using the real archive endpoint and quota-accounted GitHub transport. With 203 remaining and a 200 reserve, a ten-call hydration reaches upstream exactly three times; the fourth attempt is refused locally, the quota pool remains at 200, and SQLite records retryable interrupted work.

Focused packages, `make test-short`, lint, commit hooks, and push hooks pass.

wesm:
Addressed the remaining medium finding in `1c6af864`.

- Provider-paced archive attempts now use a separate counter keyed to the selected conservative REST/GraphQL reset. Local hourly rollover and an earlier individual-pool reset no longer replenish that allowance.
- The counter starts fresh only when the selected provider pacing reset advances. Local conditional-request refunds remain independent so late responses cannot create provider pacing credit across windows.
- Added regressions for both a local rollover inside an active provider window and staggered REST/GraphQL resets where the earlier pool invokes the budget reset callback.

The focused red/green regression, full `internal/github` package, `make test-short`, lint, commit hooks, and push hooks pass. Deployed `1c6af864` to the laptop LaunchAgent with the existing config preserved; the live API reports sync running, a 3,000 local ceiling, and GitHub sync not paused.

wesm:
Addressed the medium finding in `dd2b6fe9`.

- GitHub archive admission now installs a provider-aware allowance. Before every REST or GraphQL wire attempt, the quota transport rechecks every required live pool and refuses I/O when any pool reaches the 200-unit reserve.
- Each quota pool tracks the largest observed same-window `Remaining` delta. The next attempt reserves that cost rather than assuming one unit, and a larger actual response delta debits the in-flight allowance before another attempt can start.
- A provider reset starts cost observation at one but does not enlarge an already admitted hydration allowance.
- Added focused coverage for multi-unit response costs, reset advancement, and concurrent usage exhausting the other required resource. The API-to-SQLite e2e now models a two-unit response: the old code made three calls and fell from 203 to 197; the fixed path makes one call and stops at 201.

`make test-short` passed all 5,576 tests, lint reported zero issues, the focused allowance tests passed under `-race`, and commit/push hooks passed. Deployed `dd2b6fe9` to the laptop LaunchAgent with the existing 3,000 ceiling preserved; the live API reports the service active and GitHub sync unpaused.

wesm:
Addressed both accounting gaps in 4e58589d.

- The provider pacing epoch now retains the full REST/GraphQL reset vector. Prior provider spend is cleared only after every resource represented by the previous epoch has advanced, so an earlier REST-only rollover cannot regrant capacity while GraphQL remains in its old window.
- Persistent provider spend now records quota units independently from local wire-call spend. Each request persists its reserved cost, then reconciles any larger observed header delta; subsequent admissions therefore retain learned multi-unit costs even when a response omits rate-limit headers.

Regression coverage verifies that advancing only the earlier REST reset retains 100 units of prior spend, and that two separate admissions with a learned cost of 2 reduce paced capacity from 1,200 to 1,198 and then 1,196.

Validation:
- go test ./internal/github ./internal/server/e2etest -count=1
- make test-short (5,579 passed, 162 skipped)
- make lint (0 issues)
- go test -race ./internal/github -run "^(TestSyncBudgetProviderArchivePacingWaitsForEveryPoolReset|TestArchiveProviderQuotaCostPersistsAcrossAdmissions|TestArchiveProviderAttemptAllowance)" -count=1

Deployed to the laptop and verified the running version reports 4e58589d. Live policy reports reserve_buffer=200, local limit=3000, sync_paused=false.

wesm:
Addressed both findings in 8692c3b0.

- Archive attempts now atomically reserve effective provider headroom by credential, resource, and reset window. A headerless response leaves that reservation deducted, so a second admission at Remaining=201 cannot reuse the one unit above the 200-unit reserve. A current response header reconciles its reservation; a resource-window rollover expires old reservations.
- Persistent provider pacing spend is now keyed by resource and reset window instead of one aggregate epoch. When REST rolls and spends in its new window before GraphQL rolls, the later GraphQL reset clears only old GraphQL spend and retains the newer REST charges.

Regression coverage exercises repeated headerless admissions through the real quota and budget transports, verifies current headers release their reservation, and spends 100 old-window GraphQL units plus 25 new-window REST units across staggered resets.

Validation:
- go test ./internal/github ./internal/server/e2etest -shuffle=on
- make test-short (5,581 passed, 162 skipped)
- make lint (0 issues)
- focused race tests for both regressions and the existing multi-unit admission cases

Deployed to the laptop and verified the service reports 8692c3b0 with reserve_buffer=200, local limit=3000, and sync_paused=false.

wesm:
Addressed in f9408cda.

When an archive response advances to a newer quota reset, reconciliation now uses `Limit - Remaining` from that first new-window header as a conservative upper bound on the request cost. That bound is applied consistently to:

- persistent pacing spend in the new resource window;
- the request-local attempt allowance; and
- the new quota pool’s learned `AttemptCost`, so the next admission reserves enough live headroom.

The transport regression starts in the prior GraphQL window, returns the first new-window header with four units used and only one unit above the 200-unit reserve, and verifies that a second admission is refused without another provider call. Before the fix, `AttemptCost` stayed at one and the second request reached the wire.

Validation:
- go test ./internal/github ./internal/server/e2etest -shuffle=on
- make test-short (5,582 passed, 162 skipped)
- make lint (0 issues)
- focused race run covering reset rollover, headerless reservations, and persistent multi-unit costs

Deployed to the laptop and verified the service reports f9408cda with reserve_buffer=200, local limit=3000, and sync_paused=false.

PR #773, opened 2026-07-28
Make diff file paths copyable

wesm:
Diff file paths in the main viewer were display-only, making it cumbersome to reuse a path while reviewing a change.

- Click a file path in the diff title bar to copy the exact current path with transient feedback.
- Keep collapse and expand as a separate accessible control with explicit state and a practical hit target.


mariusvniekerk:
recovering this.  the security reviewers kinda made this one go off the rails.

PR #775, opened 2026-07-29
Reset archive state for recreated repositories

wesm:
A repository path can survive deletion and recreation while its stable provider identity changes. Retaining the previous snapshots in that case mixes distinct repository incarnations and can corrupt archive reports when item numbers are reused.

- Rebuild provider-owned state when a repository identity changes at an existing path while preserving its local project association.
- Reject pull-request-shaped responses at the issue normalization boundary.

Validated with the focused database and provider test suites, the context structural check, and branch diff checks.


wesm:
abandoning, i'll start over

PR #776, opened 2026-07-29
Use xterm's clipboard addon for tmux copy

mariusvniekerk:
Use xterm's maintained clipboard addon for tmux OSC 52 copy.

- Forward terminal input unchanged and keep xterm as the only renderer.
- Try the browser clipboard first, then use a write-only local host fallback when Zen or Firefox rejects the asynchronous write.
- Preserve real tmux copy across terminal reconnects without custom OSC parsing, gesture tracking, or mouse filtering.

The provider boundary follows VS Code desktop; code-server adds no separate host bridge. A real tmux copy was manually verified in Zen against the macOS clipboard.


mariusvniekerk:
This probably not needed anymore?

PR #780, opened 2026-07-29
Keep keyboard focus in a pooled terminal across a pane move

mariusvniekerk:
Since #766, moving a session pane (drag, split, or join) reparents its live terminal through the pool's `display:none` parking node, which blurs xterm's helper textarea; focus came back only for an explicit Focus Terminal request. Keyboard-driven tmux copy/paste stopped reaching the terminal after any pane move until the user clicked it — easy to misread as an OSC 52 clipboard failure, and invisible to the Firefox clipboard e2e tests because they click the terminal before every operation.

- The terminal pool now restores focus it took during a reparent when the wrapper held it; closing a pane still drops that intent.
- Regression test covers focus staying in the terminal across a slot move with no extra click.

🤖 Generated with Claude Code (claude-fable-5)

### DIFF
--- a/context/ui-interaction-contracts.md
+++ b/context/ui-interaction-contracts.md
@@ -302,12 +302,15 @@ Keyboard handlers must have one clear owner for each key press.
   websocket (`frontend/src/lib/stores/session-host.svelte.ts`).
 - A pooled terminal constructs immediately, even in parking, so every mounted
   session keeps its websocket; it opts out of renderer autofocus. After
-  attachment the pool focuses explicit requests and restores focus-event-tracked
-  keyboard ownership: a real pane move rips the focused textarea out of the DOM
-  silently (slot teardown fires no focusout), so never sample activeElement at
-  park time. Ownership is revoked by any other element's focus claim, and a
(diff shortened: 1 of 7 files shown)


mariusvniekerk:
The finding was confirmed and the fix redesigned in 56f0b4dfa. Park-time `activeElement` sampling can never work on the production path: real pane moves tear down the source `SessionTerminalSlot`, which removes the focused textarea from the DOM (no `focusout` fires) before the pool's effects run — verified with a real-Firefox harness against a live tmux workspace, where the original fix still lost focus after a keyboard "Split pane right".

The pool now tracks keyboard ownership through document focus events (silent DOM loss preserves it, any other element's focus claim revokes it) and restores at attachment only into unclaimed focus. Coverage: real slot-transfer Vitest tests in both template orders plus two no-stealing tests (all red on the old implementation), and `terminal-focus-pane-move.spec.ts` — promotion, keyboard pane split, post-move typing and OSC 52 copy through `page.keyboard` with no post-move click, green on chromium and firefox.

PR #783, opened 2026-07-29
Fix terminal keyboard and resize handling

mariusvniekerk:
- Keep Escape, function keys, and modified shortcuts inside focused terminals instead of triggering app-level commands.
- Resize every painted terminal from its own region, including unfocused split leaves and vertically resized dock panes.
- Keep parked terminals from overriding the size of the visible tmux pane.

[Terminal resize video](https://github.com/user-attachments/assets/627c1607-e994-47b2-85b3-a6ad7b969266)


mariusvniekerk:
[Terminal resize video](https://github.com/user-attachments/assets/627c1607-e994-47b2-85b3-a6ad7b969266)

PR #786, opened 2026-07-29
Launch sessions from the workspace header

mariusvniekerk:
Launching another session from an embedded workspace currently requires an extra trip through the controls popover, which is especially awkward when the terminal is collapsed.

- Put a direct Launch session action beside Delete for ready workspaces.
- Keep the popover entry for promoted session panes and preserve blocked and non-ready states.

Workspace header with Launch session beside Delete


mariusvniekerk:
Workspace header with Launch session beside Delete

PR #787, opened 2026-07-29
Make background daemon startup idempotent

cpcloud:
- Add `middleman start --background`, reusing a compatible proof-verified runtime or starting and verifying one.
- Publish standard daemon discovery and readiness identity from config home, including resolved runtime metadata.
- Prove the recorded endpoint owns the startup token before sending bearer credentials.
- Allow direct loopback clients with the startup-bound bearer while preserving reverse-proxy Host validation and optional API authentication.
- Keep raw foreground duplicate-start behavior unchanged and use loopback TCP across platforms.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -175,6 +175,7 @@ Coverage of real behavior is non-negotiable; the lane is chosen by the behavior 
 - No emojis in code or output
 - For database schema changes, follow `context/db-migrations.md`; `internal/db/migrations/` is the source of truth for schema evolution.
 - For HTTP API error envelopes and frontend error branching, follow `context/error-handling.md`; branch on stable codes/details rather than prose.
+- For server discovery, background startup, and daemon identity, follow `context/daemon-lifecycle.md`.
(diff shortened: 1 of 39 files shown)


cpcloud:
On the two medium findings we are not adopting:

- Restoring `X-Forwarded-Host` would violate the direct-client trust boundary. A direct request qualifies only with the valid daemon bearer, exact loopback listener authority, loopback peer, and no forwarding headers. Fabricating a forwarding header would make the CLI impersonate the configured reverse proxy. Genuine proxy traffic remains supported, while `start --background` deliberately rejects `trust_reverse_proxy=true` with authentication disabled because that configuration cannot authenticate a direct lifecycle probe.
- Discovery publication remains fatal because the config-home runtime record is a required identity surface, not auxiliary status. Allowing foreground startup without it would create a running but undiscoverable daemon: generic readers would miss it and a later `start --background` could not reuse it. Publication could be preflighted earlier for a cleaner error, but making it non-fatal would break the lifecycle contract.

The missing-directory-beneath-a-symlink finding is fixed in `93221d74` by resolving the longest existing ancestor before appending missing path components.

PR #789, opened 2026-07-30
Preserve repository history across path reuse

wesm:
- Preserve repository history when a stable provider repository ID moves to a new path.
- Treat path reuse by a different provider ID as a separate repository incarnation.
- Scope repository-owned sync, archive, notification, workspace, and clone state to that incarnation.
- Preserve exact configuration routing and settings behavior across renames and offline restarts.


wesm:
I'm closing this so we can start fresh

PR #793, opened 2026-07-30
Keep launcher closed during inline workspace deletion

wesm:
Deleting an inline workspace could briefly show the Launch Session dialog while it was being removed.

- Keep the launcher closed throughout normal and force-delete flows.
- Preserve automatic launch prompts for genuinely empty ready workspaces.

[Inline workspace deletion video](https://github.com/user-attachments/assets/e6a1a228-fbc3-4c75-880e-1fc956eba576)


wesm:
[Inline workspace deletion video](https://github.com/user-attachments/assets/e6a1a228-fbc3-4c75-880e-1fc956eba576)

PR #797, opened 2026-07-31
Keep launcher closed during explicit agent startup

wesm:
Selecting an agent from the **Create Workspace** split menu is already a complete launch decision. During startup, a newly ready workspace can temporarily report zero runtime sessions; the empty-pane fallback treated that transient state as genuinely empty and briefly displayed **Launch a session**, suggesting that a second choice was required.

This keeps the automatic fallback suppressed while the explicit intent is queued or its agent launch is in flight. If launch fails and the workspace remains empty, the normal launcher becomes available again. Regression coverage holds the launch request open to protect this transient state, and the UI interaction context now records the broader invariant.

### DIFF
--- a/context/ui-interaction-contracts.md
+++ b/context/ui-interaction-contracts.md
@@ -90,12 +90,15 @@ Interactive surfaces must agree on which item is selected.
   begun controller-less must surface on an inline surface after a layout
   switch, where no recordCreated override ever ran
   (`frontend/src/lib/stores/workspace-host.svelte.ts::effectiveRef`).
-- An explicit create-and-launch intent must be published before presentation
-  liveness checks and remain reactive after a workspace is already ready; layout
-  churn or branch reuse must not silently discard the command
+- An explicit create-and-launch intent is keyed by `(workspaceHostKey, workspaceId)`
+  and must be published before presentation liveness checks; ID-only state collides
+  local and fleet workspaces, while layout churn must not discard the command
   (`packages/ui/src/components/workspace/WorkspaceCreateSplitButton.svelte::selectTarget`,
   `packages/ui/src/stores/workspace-create-pending.svelte.ts::queueWorkspaceLaunch`,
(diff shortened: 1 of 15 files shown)


wesm:
Addressed in 2f1ed3cd2. The launcher guard now treats identity-scoped workspace creation as startup, covering the interval before the workspace-ID launch intent is queued. The full-stack test holds the real session-launch POST and uses a MutationObserver to prove the launcher never mounts while startup is pending. Verified with the full Vitest unit lane (3,383 passed, 2 skipped), the full affected Playwright file (3 passed), and the frontend check.

wesm:
Addressed in 5dc197c9b. The full-stack test now lets the real workspace-create request complete server-side, withholds its response, waits for the ready workspace pane, and checks the MutationObserver before releasing the response. Removing the `isWorkspaceCreatePending` guard makes that new pre-response assertion fail. The stronger test also exposed a cross-view handoff gap after response release, so a claimed launch intent now remains globally pending until the request settles while only one view can own the launch. Verified with frontend checks, the full unit lane (3,384 passed, 2 skipped), five repeated full-stack regression runs, and the complete affected Playwright file (3 passed).

wesm:
Addressed both ownership findings in f2bd03d1e. Claimed launch intents now carry an opaque ownership token: only the matching claim can complete them, while sibling views may discard only unclaimed intents. The terminal view settles the captured claim rather than reading the current reactive workspace ID. Added regressions for sibling discard protection, stale claim replacement, and A-to-B navigation while A launches. Verified with 3,386 unit tests, static checks, and all 3 full-stack create-and-launch Playwright scenarios.

wesm:
Addressed the host-collision finding in eca01c3a. Launch intents are now keyed by both workspace host and workspace ID, and the host argument is required across queue, lookup, claim, discard, and completion boundaries. Added store coverage for independent local/fleet intents sharing an ID and component coverage proving a local intent neither launches nor suppresses the launcher in the same-ID fleet view. Verified with 3,388 unit tests, 263 focused affected tests, static checks, and all 3 full-stack create-and-launch Playwright scenarios.

PR #802, opened 2026-07-31
Fix Codex terminal arrow-key redraws

mariusvniekerk:
- Enable enhanced terminal keyboard reporting so Codex approval menus handle arrow keys consistently.
- Align xterm core and WebGL versions while preserving accelerated glyph rendering.


wesm:
looking

PR #808, opened 2026-08-01
Require selection movement before tmux clipboard writes

wesm:
A terminal focus click can emit a one-cell tmux OSC 52 copy and replace text just copied from PR details.

- Require four CSS pixels of pointer movement before a terminal gesture can authorize browser or local clipboard writes.
- Keep outside-copy revocation while preserving keyboard and deliberate tmux drag copies.
- Cover both the PR comment focus-click path and delayed outside-copy revocation through real tmux.


mariusvniekerk:
yeah, i remember this issue.

wesm:
IT WORKS

PR #810, opened 2026-08-01
Preserve Activity filters across tab reloads

wesm:
Activity filter choices could reset after the browser reloaded or restored a discarded tab while another mode was open, because the prior Activity URL existed only in memory.

- Keep the last Activity URL for the current browser-tab session so disabled item and event filters return unchanged.
- Ignore invalid cached routes and preserve ordinary navigation when browser storage is unavailable.


wesm:
not ready yet

wesm:
confirmed working correctly

wesm:
confirmed working

PR #811, opened 2026-08-01
Preserve repository history across route reuse

wesm:
Repository names are mutable and reusable, but the database currently treats a route as identity. That can combine historical activity from distinct provider repositories or break continuity when a repository is renamed.

This establishes stable provider repository IDs as canonical identity, records route occupancy separately, and makes route reuse retire the prior occupant without moving or deleting its issues, merge requests, or archive rows. Existing canonical archives migrate in place, while blank-ID legacy rows remain inactive instead of being guessed from a matching route.

Current provider-ID writers and archive enrollment now use the same reconciliation transaction, and catalog reads support both active-route lookup and immutable-ID access to historical repositories. The schema change is contained in one migration.


wesm:
this is the new try at #789

wesm:
Stopping the fix loop here and landing this PR. Seven review rounds have been addressed on this branch; the remaining five findings from the latest round are hardening of pre-existing architecture rather than regressions in this diff, and each is now tracked as a kata in the forge project:

- **kata forge#pcry** — Bind projects to stable repository identity (schema + API change to the projects feature).
- **kata forge#qr1f** — Close the push/pull TOCTOU window against route reconciliation. Note: the prescribed fix (holding the reconciliation read lock across a network git operation) conflicts with the design decision made earlier in this PR for clone — network operations can take minutes and would stall all catalog writes — so the kata evaluates bounded alternatives instead.
- **kata forge#kyq0** — Verify credential alias identity before privileged mutations (fail-closed write-path resolution).
- **kata forge#h6rt** — Account for identity-lookup cost in optional-work admission.
- **kata forge#znm4** — Make feature-cooldown sibling keys update atomically (bounded impact: one redundant probe call).

What this PR already fences on the route-reuse threat: workspace creation, setup (with post-clone re-validation), parent and child sync snapshots, feature cooldowns, credential alias registration/clearing, startup alias registration, and workspace push/pull/branch-inspection network entry points — all with route-replacement race coverage.

PR #813, opened 2026-08-01
Preserve provider snapshots when reconciling renamed repositories

wesm:
A repository's owner/name is a mutable route: renames move it, and a vacated path can be reused by a different repository. Because sync keyed repository data by path, a rename could lose merge settings, and a reused path could mix items, notifications, and cached state from two different repositories.

This PR keys repository-owned data by the provider's stable repository ID and checks route ownership before committing async provider data:

- **Route fences.** Routes carry a generation that increments when a path changes owners. Provider-derived writes (item snapshots, notifications, watermarks, ETags, deferred comment refreshes, queued read acknowledgements) commit only while the fence captured at sync start still holds.
- **Settings.** All sync paths persist repository metadata and merge settings through one watermark-gated writer. Stale snapshots are dropped instead of overwriting newer ones, provider-omitted fields don't erase stored values, and direct PR/issue syncs persist verified settings so a repository row created for a reused path never advertises the default (all-allowed) merge availability.
- **Clones.** Sync and diff clones are stored per stable repository ID, so repositories that ever share a path never share git data. The shared fetch validates route ownership before releasing waiters; workspace clones stay path-scoped.
- **Notifications.** Sync verifies repository identity before listing, and a vacated path drops its notifications, ETags, and sync watermarks so the next occupant starts clean.

Most of the added lines are tests; the context docs record the new invariants.

### DIFF
--- a/cmd/e2e-server/main.go
+++ b/cmd/e2e-server/main.go
@@ -1496,25 +1496,35 @@ func buildAppState(
 	rootHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		if r.Method == http.MethodPost &&
 			r.URL.Path == "/__e2e/issue-workspace/reused-branch" {
-			clonePath, err := diffRepo.Manager.ClonePath(
+			identityClonePath, err := diffRepo.Manager.ClonePathForContext(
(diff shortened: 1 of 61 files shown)


wesm:
This has gotten a little out of hand, trying to reign in the complexity

PR #814, opened 2026-08-01
Keep tmux aligned with fitted terminal dimensions

wesm:
Bottom terminals could occasionally show tmux's status line one row above the panel edge. The browser used a fit preflight size even though xterm measures again while applying fit; a panel settling across a cell boundary could therefore leave xterm and the PTY disagreeing about row count.

This makes the preflight measurement authority-only and reports post-fit dimensions for both normal resizes and replay reconciliation. The regression tests model a row-boundary change in both paths.

Verification note: the full frontend unit suite passes with reduced concurrency. The default-concurrency run hit four unrelated host-load timeouts, and each affected test passed independently.


wesm:
Tested and confirmed the bug I was seeing was fixed

PR #815, opened 2026-08-02
Keep Activity PR details aligned with notification activity

wesm:
## Why

Activity can learn about new pull-request work before the corresponding PR timeline has been hydrated. That allowed the Activity row to look current while the selected detail remained stale.

Manual QA also showed that recently changed PRs are a poor proxy for the reviews a user is actively rotating through. Self-authored activity may not create a notification, and the old scheduler/detail-age interaction allowed a viewed PR comment to take 2m51s to appear.

## What changed

- Persist the last 10 unique open PR details successfully viewed. This hot set survives daemon restarts and a repeated view moves the PR to the front.
- Refresh hot PRs on the configured watched cadence, which defaults to two minutes, even when they are outside the activity window or have no notification.
- Refresh other recently active open PRs on a 10-minute warm cadence. Linked notification activity can promote a PR into this warm set without overwriting authoritative `last_activity_at`.
- Remove a hot PR immediately when a persisted provider snapshot marks it closed or merged. Terminal details cannot re-enter the hot set, and deleted PR rows cascade out.
- Serialize scheduled and immediate watched-PR passes so they cannot race on host cadence state or duplicate provider work.
- Preserve one pending rerun when newer detail-viewing intent arrives during an in-flight sync.
- Broadcast `data_changed` after watched detail persistence so a selected Activity detail reloads from SQLite.

## API cost controls

Hot refresh is bounded to 10 persisted PRs. Warm PRs now refresh every 10 minutes instead of five. The existing host cadence gates, repository tracking, rate-limit backoff, hourly sync budgets, and conditional GitHub detail requests remain in force; unchanged GitHub PR resources can still return 304.

## QA expectations

- Opening a PR detail makes it hot. Provider-side comments or reviews should appear by the next two-minute watched pass, often sooner through the foreground detail sync.
- Rotating among up to 10 open PRs keeps all of them hot; opening an eleventh displaces the least recently viewed PR.
- A PR disappears from the hot set as soon as sync observes it closed or merged.
- A recently active but unviewed PR is warm and may take up to 10 minutes to refresh.


wesm:
Addressed the high-severity coverage finding in `a4d7c1764`. The new full-stack Activity regression starts with an already-selected stale PR, adds a provider-only comment plus a newer linked notification, runs the real watched fast-sync selector, verifies SQLite/API persistence, and confirms the resulting `data_changed` reload exposes the comment in the open detail; the focused Chromium spec passes.

wesm:
doing some manual QA on this

wesm:
test

wesm:
working on this

wesm:
Addressed the watched-MR concurrency finding in `0650ef8ef`.

- Added a dedicated pass-level mutex so scheduled and immediate `SyncWatchedMRs` runs cannot overlap provider work or race on `nextWatchSyncAfter`.
- Added a blocking-provider regression test that reproduced `maxInFlight = 2` before the fix and now holds it at 1; it passes across 20 repetitions and under the race detector.
- The same commit implements the final freshness policy: persisted last-10 viewed PRs are hot, ordinary recently active or notification-promoted PRs are warm at 10 minutes, and terminal PRs are evicted immediately.

wesm:
Addressed in 1cf0cd564. LatestOpenPRNotificationActivity now treats a non-null notification repo_id as the stable subject identity, so cached route fields cannot orphan activity after a rename or redirect it after route reuse. Legacy null-ID rows fall back only through the current repository route. Added regression coverage for both rename and route-reuse behavior.

PR #825, opened 2026-08-03
Fix terminal click jitter overwriting the clipboard

wesm:
Ordinary terminal focus clicks can slip five pixels. The previous four-pixel absolute threshold could treat that motion as a tmux selection and replace clipboard text copied from PR details.

Selection authorization now follows readable rendered cell geometry while preserving the four-pixel floor, so realistic focus jitter remains unauthorized and deliberate drag-copy still works.


mariusvniekerk (approved):

mariusvniekerk:
> At cell widths below 3px

WHICH NO HUMAN CAN SEE YOU IDIOT ROBOT....

wesm:
The requested greater-than-five-pixel absolute floor is not being adopted. It addresses configured cell widths below three pixels, which are not a readable terminal geometry and are outside the interaction behavior we want to optimize.

The regression now uses default geometry, asserts the rendered cell width exceeds five pixels, and applies a trusted five-pixel focus-click movement. It fails with the old four-pixel-only predicate and passes with the current cell-relative gate. Existing real-tmux drag-copy coverage remains.

PR #826, opened 2026-08-03
fix: stabilize hosted Go CI fallbacks

mariusvniekerk:
## Why

PR #823 restored Forge's managed-runner routing, but pull-request dispatch intentionally invokes the reusable workflow from `main`. Until a repair reaches `main`, PR checks therefore bootstrap on the hosted lane. Those runners can expose a rustup `cargo` shim without an installed toolchain, and the lower worker profile made two timing-based test assumptions unreliable.

The investigation also exposed a real cleanup race: a newly created tmux backend could outlive a failed metadata write if request cancellation or natural attachment exit removed the in-memory session first.

## What changed

- Require `cargo --version` to succeed before running Rust-backed Go E2Es.
- Synchronize the existing-worktree replacement test at the exact repository-lock boundary instead of sleeping for 80 ms.
- Make launch rollback cancellation-independent and clean the tmux backend by the launch result even if the in-memory session has already exited.
- Document the runtime cleanup invariant.

## Verification

- All commit and push hooks passed, including Go short tests, golangci-lint, nilaway, and repository guardrails.
- `scripts/context-sync --check` and branch diff integrity passed.
- The worktree lock regression passed 10 repeated race-instrumented runs.
- The Rust manager and rollback server regressions passed five repeated race-instrumented runs.
- Both Rust workspace E2Es passed.

During broader diagnosis, a complete `internal/server -race` run also surfaced one unrelated pre-existing failure in `TestWorkspaceRefreshProceedsThroughIssueScopePartialSyncFailure`; the regressions changed here pass independently and repeatedly.

Origin: Buzz channel `23c2f26c-fce2-5341-b4c1-f149e7cbbfea`, thread `674b32e20376aa7f86aafd2e4069328e03c8a198f1d6b7b5f622b58c66362d19`.


wesm:
looking

PR #828, opened 2026-08-03
feat: show Roborev configuration on repository cards

mariusvniekerk:
Repository cards currently provide no way to tell whether Roborev will review commits. This adds a hook-backed indicator using the official Roborev mark while keeping the optional daemon and filesystem probe separate from core repository summaries.

- Cache successful inventory and definitive hook results for the Kenn Forge process lifetime.
- Preserve provider-aware repository matching and expose whether partial results are complete.
- Show the branded indicator only when a matching checkout has an executable Roborev-managed hook.


mariusvniekerk:
Repository card with the grayscale, transparent Roborev indicator:

PR #829, opened 2026-08-03
docs(terminal): record snapshot accelerator no-go

wesm:
- Record the production-path finding: workspace switching reparents pooled runtime-session terminals rather than opening the primary terminal endpoint.
- Keep runtime replay-boundary eligibility explicit while primary and Fleet terminals retain raw attachment behavior.
- Remove the unused snapshot/handover prototype and its rejected implementation spec; the concise lifecycle invariant and PR discussion preserve the actionable lesson.

### DIFF
--- a/context/workspace-runtime-lifecycle.md
+++ b/context/workspace-runtime-lifecycle.md
@@ -23,7 +23,9 @@ Kenn Forge manages three related but different things:
 
 Rules:
 
-- The base workspace `tmux` tab is part of the durable workspace experience.
+- The base workspace tmux session remains durable backend state, but production
+  workspace terminal panes are pooled runtime sessions; do not treat the primary
+  terminal endpoint as a mounted SPA workflow (`frontend/src/lib/components/terminal/PooledSessionTerminal.svelte`).
 - Launched agent sessions and shell sessions are not durable after natural exit.
 - The shell drawer is a singleton per workspace, but a tmux-backed shell should
   survive kenn-forge server restarts until the shell exits or the workspace is
--- a/frontend/src/lib/components/terminal/XtermTerminalPane.svelte
+++ b/frontend/src/lib/components/terminal/XtermTerminalPane.svelte
@@ -27,6 +27,7 @@
     primaryTerminalFontFamily,
   } from "./terminalFontFamily.js";
   import { createInitialFocusIntent } from "./terminal-focus.js";
+  import { supportsReplayBoundary } from "./terminalReplayBoundary.js";
   import {
     clearSharedTerminalTextureAtlas,
     registerTerminalTextureAtlasParticipant,
@@ -399,19 +400,6 @@
     return `${proto}://${location.host}${withBasePath(withConnectionParams)}`;
   }
 
-  function supportsReplayBoundary(): boolean {
-    if (!websocketPath) return false;
-    try {
-      const pathname = new URL(websocketPath, window.location.href).pathname;
-      return (
-        !pathname.includes("/fleet/hosts/") &&
-        /\/workspaces\/[^/]+\/runtime\/sessions\/[^/]+\/terminal$/.test(pathname)
-      );
-    } catch {
-      return false;
-    }
-  }
-
   function withBasePath(path: string): string {
     const normalizedPath = path.startsWith("/") ? path : `/${path}`;
     if (!basePath) return normalizedPath;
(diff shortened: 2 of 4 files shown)


wesm:
Addressed the four RoboRev findings in the force-pushed single commit `6d82d7b7e`:

- provisional snapshots now require a `replay_boundary`-aware client, preventing Fleet/raw scrollback contamination;
- staged tmux attaches drain final PTY output before the exit frame, while handler-owned cancellation preserves exit delivery;
- profiling records interactive readiness from the xterm boundary-write callback used by production;
- variant selection rejects process-limit rejections and requires both idle and contention conditions to pass.

Regression coverage includes boundary gating, fast-exit output ordering, browser parser readiness, and all-condition scoring. Post-rebase Go vet/tests/race, frontend typecheck/focused tests, and the 100-workspace profile gate pass. The profile selected one-shot with zero misses, failures, limiter rejections, or input discards (p95 first paint: 23.1 ms idle, 37.1 ms contention).

wesm:
Addressed the three staged-bridge findings in ad8eba044:

- Wrapped tmux commands now capture the session client set before launch and discover the unique new client afterward, with a real child-spawning wrapper integration test.
- Buffer-bound promotion includes the entire overflowing PTY read in one authoritative WebSocket frame before replay_ready, covered across a split UTF-8 rune and CSI sequence.
- Any PTY read termination now preserves received bytes and waits boundedly for Command.Wait before reporting an exit code, covered by a real-PTY fast-exit test.

Verified with the terminal/tmux and workspace integration suites, go test -race, go vet, golangci-lint, NilAway, frontend checks, and no-TERM tmux integration runs.

wesm:
Addressed both RoboRev findings in `45c942cf5`.

- Authoritative primary handover frames now begin with CAN before reset/output, so provisional parser state cannot leak across the boundary. Primary `replay_ready` uses an empty xterm write as the parse-ordering barrier instead of canceling authoritative state, with real-browser regressions for split UTF-8 and DECSET sequences.
- Clients without `replay_boundary` now bypass barrier discovery, handover timers, and the input gate; their PTY input/output streams immediately. Backend and server E2E coverage verifies that path.

Affected terminal, tmux integration, browser, race, lint, type, and full frontend test lanes pass.

wesm:
Addressed both findings in `ee1cfa107`.

- Primary workspace terminals now suppress stdin, typed/binary input, mouse reports, and paste until the authoritative empty-write callback has parsed terminal modes. Runtime and non-boundary clients retain their existing input semantics. Unit coverage and a real-Chromium component regression delay `replay_ready` and verify multiline paste is encoded with bracketed-paste framing only afterward.
- The performance gate retains the primary transport microbenchmark and now also performs real SPA sidebar route switches, scoring the production `workspace-switch:first-paint` measure across every sample. SPA and primary transport run in separate isolated Forge processes so persistent runtime attachments cannot contaminate the constrained transport budget; both p95 targets must pass.

The profiling suite selected one-shot with zero misses, failures, process-limit rejections, or input discards. Under 28/32-slot contention, SPA first-paint p95 was 39.6 ms and primary-transport p95 was 23.1 ms. Full frontend unit/browser/static checks, the profiling Playwright suite, and the terminal Go package pass.

wesm:
What we learned:

The original premise was wrong for the production workspace-switch path. The SPA mounts pooled runtime-session terminals and reparents their existing live terminal wrappers; it does not open a fresh primary `/workspaces/{id}/terminal` attachment on each switch. The primary endpoint still exists, but the profiler mount that exercised it was E2E-only.

That means the measured 35.5 ms p95 was valid for the surrogate primary-terminal path, but it was not evidence that tmux control mode improves the workflow users actually run. Shipping the snapshot broker would therefore add capture, handover, parser-boundary, sizing, and lifecycle complexity without removing a production per-switch attach cost.

The right decision is to remove the prototype, retain the raw primary endpoint and the production runtime-session replay behavior, and keep the design as prior art. We should revisit control mode only if profiling around 100 workspaces shows that the persistent runtime-session tmux clients or PTYs themselves are the scaling bottleneck. That would be a different project—multiplexing or replacing the production runtime transport—not accelerating primary-terminal first paint.

Process takeaway: confirming that a benchmark exercises the production mount and transport must be gate zero for future performance work.

wesm:
Follow-up decision: we are not retaining the full snapshot-accelerator spec. It was an implementation-ready design for an architecture we rejected after validating the production mount path, so keeping it would be more likely to mislead than help. The squashed branch now preserves only the concise runtime-lifecycle invariant and the existing runtime-session replay behavior. This PR discussion is sufficient history; control mode should be reconsidered only with new production-path evidence.

PR #830, opened 2026-08-03
Publish kenn-forge docs through Vercel

mariusvniekerk:
kenn-forge documentation must describe the binaries users can actually download, so production deploys the tagged checkout only after release publication succeeds. The deploy job rejects tags whose commit is not reachable from protected `main` and enters the protected `docs-production` environment before Vercel credentials are available. The Vercel GitHub App is intentionally absent: previews remain on-demand and pull requests avoid bot comments and deployment-status noise.

Vercel still performs the complete pinned source build, including the real seeded e2e screenshot generation and rendered-site verification. The public site adds the canonical favicon, static zoomable workflow screenshots with the sanitized Codex transcript and interactive footer, lowercase living-document branding, and latest-release download links with safe architecture and Releases fallbacks.

The paired infrastructure change in [kenn-ops #106](https://github.com/kenn-io/kenn-ops/pull/106) must remove the Vercel Git connection, protect release tags, and own the required-approval `docs-production` environment plus its GitHub-encrypted Vercel secrets. OpenTofu apply remains operator-owned.


wesm:
> [!NOTE]
> This comment was written by Claude on behalf of the maintainer.

The build pipeline here looks right, but I'd like to change the deployment trigger before we wire this up. Two issues with the Git-triggered model as designed:

1. It deploys production on every push to `main`, so forge.kenn.io would track main's tip rather than the latest release. Since Forge ships as tagged binaries via `release.yml`, the site would regularly document features and flags that don't exist in the version users can actually install.
2. It depends on installing the Vercel GitHub App, which we don't want — its PR comments and deployment statuses are too noisy. Without the App, Vercel never reacts to Git pushes at all, so the automatic preview and production deploys described in the design doc won't happen.

Proposal: keep the build pipeline exactly as this PR implements it, and deploy via the Vercel CLI at release time instead:

- Add a `deploy-docs` job to `release.yml`, gated on the release job, that runs `vercel deploy --prod` with `VERCEL_TOKEN`, `VERCEL_ORG_ID`, and `VERCEL_PROJECT_ID` secrets. This is a source deployment: the CLI uploads the tagged checkout and Vercel runs the same install/build/verification commands this PR defines.
- Previews become on-demand via the existing `make docs-deploy-staging` target rather than automatic per-PR.

Result: forge.kenn.io updates exactly when a release is cut, no Vercel bot activity on PRs, and docs never describe unreleased features. Docs fixes merged between releases won't reach the live site until the next tag, though a manual `make docs-deploy` from a tagged checkout stays available as an escape hatch.

This does put a Vercel token in GitHub secrets, which the design doc set out to avoid — but that constraint was only achievable via the Git App integration. With the App excluded, a CLI deploy from Actions is the only automated path; the token should be scoped as narrowly as Vercel allows. The design doc's Git-integration and rollout sections need rewriting to match.

PR #836, opened 2026-08-05
Pace archive hydration off provider quota with a reserve floor

wesm:
Archive hydration was throttled by two mechanisms that duplicate the live rate-limit enforcement forge already has. Both regularly blocked hydration while GitHub reported thousands of requests remaining.

## What changed

- When a credential's REST and GraphQL pools are known, archive admission uses provider quota directly: available spend is remaining minus a reserve, where **each pool keeps its own reserve of max(limit/5, 200)** for live sync.
- The quadratic pacing curve is gone. A fresh quota window grants full archive throughput immediately instead of ramping up over the hour.
- Reserves are enforced per pool, at admission and on every wire attempt, so a 15,000-limit pool keeps a 3,000 floor even when a 5,000-limit pool is the smallest.
- When admission defers on the reserve, it retries when the deficient pool resets, not the latest reset across all pools.
- Archive attempts covered by a quota reservation no longer count against `sync_budget_per_hour` — that knob meters live sync only. Providers without rate-limit headers (Gitea/Forgejo) and requests without a reservation keep the existing local-budget pacing.
- New endpoint `GET /archive/pacing` reports each credential's binding pool: limit, remaining, reserve, available, and reset. Credentials with incomplete quota data appear with `known: false` instead of being omitted.

## Why

- The old curve (`surplus × elapsedFraction²`) allowed almost no archive spend early in each quota window. Observed on a live drain: 35-minute stalls with 3,520/5,000 core remaining, averaging ~24 items/hour.
- The local budget made archive and live sync compete for one knob. The default (500/hour) starves hydration entirely; draining a backlog meant repeatedly raising it and restarting the daemon.

## Operator notes

- `sync_budget_per_hour` values raised to work around the old pacing can return to the default.
- The rate-limits budget meter no longer moves with GitHub archive traffic; archive spend is visible in provider quota and `/archive/pacing`.

### DIFF
--- a/context/github-sync-invariants.md
+++ b/context/github-sync-invariants.md
@@ -270,16 +270,18 @@ fallback repository listing.
 - GitHub archive code owns historical identity inventory only; hydration must invoke ordinary item sync instead of adding archive-specific lookup, normalization, or persistence. (`internal/github/pages.go::ListIssuesPage`, `internal/github/sync.go::SyncArchiveItem`)
 - Archive item hydration bypasses persisted parent-detail ETags; an unchanged parent representation does not prove that legacy lifecycle timelines are complete. (`internal/github/sync.go::SyncArchiveItem`)
 - Archive issue hydration treats timeline failures as hard errors; ordinary issue refresh remains best-effort for that optional dataset. (`internal/github/sync.go::refreshIssueTimeline`)
-- GitHub archive hydration paces off provider quota alone: availability is
-  remaining minus the archive reserve (`max(limit/5, RateReserveBuffer)`),
-  enforced at admission and again at every wire attempt. Headerless Gitealike
-  hosts spend configured local hourly surplus above `archiveLiveFloor`.
-  Attempts covered by a registry reservation do not debit the local ceiling —
-  `sync_budget_per_hour` meters live sync only — and live work preempts the
-  archive lease.
+- GitHub archive admission with a known registry pacing window paces off
+  provider quota alone: availability is remaining minus the archive reserve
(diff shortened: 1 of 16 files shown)


wesm:
Both findings addressed as of `8eb2d3ea`:

**Per-pool reserves** — fixed in `4b220827` and `8eb2d3ea`. The pacing window now computes headroom per pool (remaining minus that pool's own max(limit/5, 200) reserve) and carries the minimum; admission, the per-attempt reservation, and `/archive/pacing` all consume it. Reserve denials retry at the deficient pool's reset, the endpoint reports the binding pool's own numbers, and an e2e case with unequal limits (15,000-limit REST at its 3,000 floor, 5,000-limit GraphQL with headroom) verifies the full path: no upstream call, persisted `budget_exhausted` deferral, retry at the deficient pool's reset. The mutation check (reverting to a flat 200 buffer) fails the new e2e.

**Archive bypassing `sync_budget_per_hour`** — this is the intended design of the PR, not an oversight, so not changing it:

- The per-pool 20% reserve is the protection for shared quota. Archive can never take any pool below max(limit/5, 200), so a standard token always keeps 1,000+ requests/hour for maintainer operations and other consumers — enforced from live headers at every wire attempt, which is stronger than a local counter that drifts from real quota.
- Coupling archive to `sync_budget_per_hour` is the failure mode this PR removes: at the default of 500 the coupling starves hydration completely even when quota is idle, and any value large enough to drain an archive stops meaningfully metering live sync.
- A separate archive ceiling is a knob with no current user. If a deployment needs a larger floor for co-tenant consumers, making the reserve fraction configurable is the natural extension.

PR #839, opened 2026-08-05
Support archived repos as archive-only and harden archive seeding

wesm:
## Problem

A provider-archived repository cannot be configured: resolution rejects it, so its historical data can never be collected. Worse, an archived repo in config crash-loops the daemon at startup, and the error names no repository.

## Changes

- Accept archived repositories (exact and glob, GitHub and GitLab) and track them archive-only: live sync, notifications, and fast sync skip them; archive collection works normally.
- Detect archived flips without a restart: settings add/refresh applies fresh metadata, each sync pass re-checks archived refs (an unarchived repo resumes live syncing; a repo found archived mid-pass stops before syncing content), and throttled credential buckets defer the refresh.
- Degrade seeding per repository: a bad config entry is logged with its identity and skipped instead of aborting startup, and the archive worker skips unresolvable refs instead of starving healthy repositories.
- Track renamed repositories once: dedupe the tracked set by stable provider id, record each exact entry's config path on its resolved ref, and use that provenance in reload fallback, startup catalog recovery (including credential aliases), settings merges, entry removal, and sync publications — a rename never duplicates a repo, loses its archived flag, or leaks state to a route successor.
- Defer removal pausing while any configured ref cannot be matched to a repository row, so an incomplete protection list cannot pause the wrong repository's archive.
- Fix a pre-existing data race between the config watcher and a settings test on `cfgPath`.

## Notes

- While an unresolvable config entry persists, removal pausing stays deferred; the recurring warning names the refs to fix.
- Startup rename recovery binds a configured path to its current catalog occupant, else a sole historical match; a stale binding self-corrects on the first successful resolve.

### DIFF
--- a/cmd/kenn-forge/main.go
+++ b/cmd/kenn-forge/main.go
@@ -763,6 +763,14 @@ func resolveStartupRepos(
 				expanded = fallbackExactFromDB(ctx, database, raw)
 				if len(expanded) == 0 {
 					expanded = ghclient.FallbackConfiguredRepoRefs(nil, raw)
+				} else {
+					// The catalog recovered a stable identity on a renamed
+					// route; without the alias, repo-scoped credentials for
+					// the configured route would fall through to owner or
+					// host credentials.
+					ghclient.RegisterConfiguredRepoCredentialAliases(
+						githubRouters, raw, expanded,
+					)
 				}
 			}
 		} else {
--- a/cmd/kenn-forge/main.go
+++ b/cmd/kenn-forge/main.go
@@ -760,7 +760,10 @@ func resolveStartupRepos(
 					ctx, database, raw,
 				)
 			} else {
-				expanded = ghclient.FallbackConfiguredRepoRefs(nil, raw)
+				expanded = fallbackExactFromDB(ctx, database, raw)
+				if len(expanded) == 0 {
+					expanded = ghclient.FallbackConfiguredRepoRefs(nil, raw)
+				}
 			}
 		} else {
 			ghclient.RegisterConfiguredRepoCredentialAliases(
@@ -784,6 +787,71 @@ func splitProviderHostKey(key string) (string, string) {
 		return key, ""
 	}
 	return platformName, host
+}
+
+// fallbackExactFromDB recovers the stored identity for an exact config
+// entry whose provider resolution failed at startup. Catalog route history
+// follows a provider-side rename, so the fallback keeps the stable provider
+// id — deduplicating against overlapping resolved entries — instead of
+// synthesizing an identity-less ref under the stale configured route.
+func fallbackExactFromDB(
+	ctx context.Context,
+	database *db.DB,
+	raw config.Repo,
+) []ghclient.RepoRef {
+	if database == nil {
+		return nil
+	}
+	repoPath := strings.TrimSpace(raw.RepoPath)
+	if repoPath == "" {
+		repoPath = raw.Owner + "/" + raw.Name
+	}
+	entries, err := database.ListRepositoryCatalog(ctx, db.RepositoryCatalogFilter{
+		Platform:     raw.PlatformOrDefault(),
+		PlatformHost: raw.PlatformHostOrDefault(),
+		RepoPath:     repoPath,
+		Lifecycle:    db.RepositoryLifecycleActive,
+	})
+	if err != nil {
+		slog.Warn("fallback exact from db", "err", err)
+		return nil
+	}
+	entry := catalogEntryForConfiguredRoute(entries, repoPath)
+	if entry == nil {
+		return nil
+	}
+	return []ghclient.RepoRef{{
+		Platform:           platform.Kind(raw.PlatformOrDefault()),
+		Owner:              entry.Repository.Owner,
+		Name:               entry.Repository.Name,
+		PlatformHost:       entry.Repository.PlatformHost,
+		RepoPath:           entry.Repository.RepoPath,
+		PlatformExternalID: entry.Repository.PlatformRepoID,
+		WebURL:             entry.Repository.WebURL,
+		CloneURL:           entry.Repository.CloneURL,
+		DefaultBranch:      entry.Repository.DefaultBranch,
+		ConfiguredRepoPath: repoPath,
+	}}
+}
+
+// catalogEntryForConfiguredRoute prefers the repository currently occupying
+// the configured route; a reused route may also historically match the
+// renamed repository that held it before. Multiple historical matches with
+// no current occupant cannot be attributed safely.
+func catalogEntryForConfiguredRoute(
+	entries []db.RepositoryCatalogEntry, repoPath string,
+) *db.RepositoryCatalogEntry {
+	for i := range entries {
+		for _, route := range entries[i].Routes {
+			if route.Current && strings.EqualFold(route.RepoPath, repoPath) {
+				return &entries[i]
+			}
+		}
+	}
+	if len(entries) == 1 {
+		return &entries[0]
+	}
+	return nil
 }
 
 // fallbackGlobFromDB returns repos from the database that match
--- a/cmd/kenn-forge/main.go
+++ b/cmd/kenn-forge/main.go
@@ -748,8 +748,7 @@ func resolveStartupRepos(
 	database *db.DB,
 	githubRouters map[string]*ghclient.HostRouter,
 ) []ghclient.RepoRef {
-	seen := make(map[string]struct{})
-	repos := make([]ghclient.RepoRef, 0, len(cfg.Repos))
+	set := ghclient.NewExpandedRepoSet()
 	for _, raw := range cfg.Repos {
 		_, expanded, err := ghclient.ResolveConfiguredRepoWithRegistry(
 			ctx, registry, raw,
@@ -769,18 +768,10 @@ func resolveStartupRepos(
 			)
 		}
 		for _, repo := range expanded {
-			key := string(repoPlatform(repo)) + "\x00" +
-				strings.ToLower(repoHost(repo)) + "\x00" +
-				strings.ToLower(repo.Owner) + "\x00" +
-				strings.ToLower(repo.Name)
-			if _, ok := seen[key]; ok {
-				continue
-			}
-			seen[key] = struct{}{}
-			repos = append(repos, repo)
+			set.Add(repo, err == nil)
 		}
 	}
-	return repos
+	return set.Refs()
 }
 
 func providerHostKey(platformName, host string) string {
@@ -793,23 +784,6 @@ func splitProviderHostKey(key string) (string, string) {
 		return key, ""
 	}
 	return platformName, host
-}
-
-func repoPlatform(repo ghclient.RepoRef) platform.Kind {
-	if repo.Platform != "" {
-		return repo.Platform
-	}
-	return platform.KindGitHub
-}
-
-func repoHost(repo ghclient.RepoRef) string {
-	if repo.PlatformHost != "" {
-		return strings.ToLower(repo.PlatformHost)
-	}
-	if host, ok := platform.DefaultHost(repoPlatform(repo)); ok {
-		return host
-	}
-	return platform.DefaultGitHubHost
 }
 
 // fallbackGlobFromDB returns repos from the database that match
--- a/cmd/kenn-forge/main_test.go
+++ b/cmd/kenn-forge/main_test.go
@@ -375,6 +375,47 @@ func TestResolveStartupReposRecoversRenamedExactEntryFromCatalog(t *testing.T) {
 	assert.Equal("acme/tools", repos[0].ConfiguredRepoPath)
 }
 
+func TestResolveStartupReposRegistersCredentialAliasForCatalogFallback(t *testing.T) {
+	require := require.New(t)
+	database := dbtest.Open(t)
+	now := time.Now().UTC()
+	before := db.GitHubRepoIdentity("github.com", "acme", "tools")
+	before.PlatformRepoID = "repo-acme-tools"
+	_, _, err := database.ReconcileRepositoryObservation(
+		t.Context(), before, now.Add(-time.Hour),
+	)
+	require.NoError(err)
+	after := db.GitHubRepoIdentity("github.com", "acme", "tools-new")
+	after.PlatformRepoID = "repo-acme-tools"
+	_, _, err = database.ReconcileRepositoryObservation(t.Context(), after, now)
+	require.NoError(err)
+
+	client := getRepoFailingClient{&testutil.FixtureClient{}}
+	router, err := ghclient.NewHostRouter("github.com", &ghclient.Route{
+		Key:    ghclient.RouteKey{Host: "github.com", Owner: "acme", Name: "tools"},
+		Client: client,
+	})
+	require.NoError(err)
+	cfg := &config.Config{Repos: []config.Repo{{Owner: "acme", Name: "tools"}}}
+
+	repos := resolveStartupRepos(
+		t.Context(),
+		cfg,
+		mustProviderRegistry(t, map[string]ghclient.Client{"github.com": client}),
+		database,
+		map[string]*ghclient.HostRouter{"github.com": router},
+	)
+
+	require.Len(repos, 1)
+	require.Equal("tools-new", repos[0].Name)
+	// The recovered renamed route must keep resolving to the exact entry's
+	// repo-scoped credential instead of falling through to owner or host
+	// routes.
+	route, err := router.RouteForRepo("acme", "tools-new")
+	require.NoError(err)
+	require.Equal("tools", route.Key.Name)
+}
+
 func TestResolveStartupReposFallsBackToDBForOfflineGlobs(t *testing.T) {
 	assert := assert.New(t)
 	require := require.New(t)
--- a/cmd/kenn-forge/main_test.go
+++ b/cmd/kenn-forge/main_test.go
@@ -249,6 +249,57 @@ func TestResolveStartupReposExpandsConfiguredGlobs(t *testing.T) {
 	}, repos)
 }
 
+type getRepoFailingClient struct {
+	*testutil.FixtureClient
+}
+
+func (getRepoFailingClient) GetRepository(
+	context.Context, string, string,
+) (*gh.Repository, error) {
+	return nil, errors.New("transient resolve failure")
+}
+
+func TestResolveStartupReposPrefersResolvedOverFallbackDuplicates(t *testing.T) {
+	assert := assert.New(t)
+	// The exact entry fails resolution and falls back to a synthetic ref;
+	// the overlapping glob resolves the same repo as archived. The resolved
+	// metadata must win or the archived repo would be polled as live.
+	cfg := &config.Config{
+		Repos: []config.Repo{
+			{Owner: "acme", Name: "archived"},
+			{Owner: "acme", Name: "*"},
+		},
+	}
+	client := getRepoFailingClient{&testutil.FixtureClient{
+		ReposByOwner: map[string][]*gh.Repository{
+			"acme": {{
+				NodeID:   new("repo-acme-archived"),
+				Name:     new("archived"),
+				Owner:    &gh.User{Login: new("acme")},
+				Archived: new(true),
+			}},
+		},
+	}}
+
+	repos := resolveStartupRepos(
+		t.Context(),
+		cfg,
+		mustProviderRegistry(t, map[string]ghclient.Client{"github.com": client}),
+		nil,
+		nil,
+	)
+
+	assert.Equal([]ghclient.RepoRef{{
+		Platform:           "github",
+		Owner:              "acme",
+		Name:               "archived",
+		PlatformHost:       "github.com",
+		RepoPath:           "acme/archived",
+		PlatformExternalID: "repo-acme-archived",
+		Archived:           true,
+	}}, repos)
+}
+
 func TestResolveStartupReposKeepsExactReposWhenResolutionFails(t *testing.T) {
 	assert := assert.New(t)
 	cfg := &config.Config{
--- a/cmd/kenn-forge/main_test.go
+++ b/cmd/kenn-forge/main_test.go
@@ -232,21 +232,19 @@ func TestResolveStartupReposExpandsConfiguredGlobs(t *testing.T) {
 
 	assert.Equal([]ghclient.RepoRef{
 		{
-			Platform:           "github",
-			Owner:              "roborev-dev",
-			Name:               "kenn-forge",
-			PlatformHost:       "github.com",
-			RepoPath:           "roborev-dev/kenn-forge",
-			ConfiguredRepoPath: "roborev-dev/*",
+			Platform:     "github",
+			Owner:        "roborev-dev",
+			Name:         "kenn-forge",
+			PlatformHost: "github.com",
+			RepoPath:     "roborev-dev/kenn-forge",
 		},
 		{
-			Platform:           "github",
-			Owner:              "roborev-dev",
-			Name:               "archived",
-			PlatformHost:       "github.com",
-			RepoPath:           "roborev-dev/archived",
-			Archived:           true,
-			ConfiguredRepoPath: "roborev-dev/*",
+			Platform:     "github",
+			Owner:        "roborev-dev",
+			Name:         "archived",
+			PlatformHost: "github.com",
+			RepoPath:     "roborev-dev/archived",
+			Archived:     true,
 		},
 	}, repos)
 }
@@ -299,7 +297,7 @@ func TestResolveStartupReposPrefersResolvedOverFallbackDuplicates(t *testing.T) 
 		RepoPath:           "acme/archived",
 		PlatformExternalID: "repo-acme-archived",
 		Archived:           true,
-		ConfiguredRepoPath: "acme/*",
+		ConfiguredRepoPath: "acme/archived",
 	}}, repos)
 }
 
--- a/cmd/kenn-forge/main_test.go
+++ b/cmd/kenn-forge/main_test.go
@@ -232,19 +232,21 @@ func TestResolveStartupReposExpandsConfiguredGlobs(t *testing.T) {
 
 	assert.Equal([]ghclient.RepoRef{
 		{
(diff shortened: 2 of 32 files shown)


wesm:
Re the combined review (7545aaa):

- **Duplicate repositories after rename** — fixed in cddaf10c. Expanded and tracked repo sets now reconcile by stable provider id when present, with route-key fallback; covered by unit tests on both dedup layers.
- **Stale sync clearing archived state** — fixed in cddaf10c. `publishResolvedRepository` preserves the currently tracked archived flag when the republication carries no fresh provider metadata; regression test included.
- **Archived repos in import preview** — intentionally unchanged in this PR. The preview feeds an add-repo flow whose UI presents repos as live sync candidates; listing archived repos there without an archived label would invite adds with surprising (archive-only) behavior. Archived repos can be added today via the config file or the add-repo API, both covered by tests. Surfacing archived repos in the preview with proper labeling is UI scope worth its own change.

wesm:
Re the commit review on cddaf10c — pushing back on all three, with reasoning:

- **Provider response racing a settings flip (`archivedAuthoritative`)**: the remaining window is a provider response in flight across the same seconds as a settings refresh. Archived here has the same last-writer-wins semantics every other provider-resolved field (default branch, URLs, rename routes) has always had, and the state self-corrects on the next sync or resolution of that repo. A tracked-set generation counter is real machinery for a self-healing race that requires sub-second interleaving to trigger; not adding it in this PR.

- **ID-less fallback duplicating a renamed route**: a synthesized fallback for a failed exact resolution has never carried a provider id, and first-wins dedup retained both routes before this PR too — this is a pre-existing rename edge, not a regression. This PR narrows it (fallbacks that carry ids — the config-reload case — now collapse). The remaining case requires a rename plus a transient resolution failure on the old route in the same pass, and the duplicate costs redundant syncs of one repo until the entry resolves, not data corruption.

- **Additional e2e coverage**: the config-reload path already has full-stack coverage with real SQLite added in 7545aaa6 (failing exact entry, overlapping glob, fallback-vs-resolved replacement, notification-lane exclusion). An e2e for the publication race would need deterministic mid-reload thread interleaving — a flake generator for a race whose unit semantics are pinned by `TestPublishResolvedRepositoryPreservesNewerArchivedState`. The renamed-route dedup is pinned at both layers by unit tests on `ExpandedRepoSet` and `mergeTrackedRepos`.

wesm:
e0d44598 addresses both findings from the combined review of cddaf10:

**High — skipped refs starved the archive worker.** `RunEligible` resolved the full configured set atomically, so a ref that seeding skipped errored every one-second pass and blocked archive work for all healthy repositories. It now resolves per-ref via `resolveRepositoriesTolerant`, dropping repository-scoped failures (debug-logged; `EnsureConfigured` already warns per reconciliation) and aborting only on context cancellation. Regression test: `TestRunEligibleSkipsUnresolvableConfiguredRef` (one healthy repo with an active archive plus one unresolvable configured ref; the healthy repo's inventory must complete). `StartAll`/`PauseAll` keep atomic resolution deliberately — they are explicit operator actions where surfacing the failing entry is more useful than silently skipping it, and they don't run in a loop.

**Medium — ID-less fallback after a rename.** Config-resolved refs now record the config-entry path they came from (`RepoRef.ConfiguredRepoPath`), preserved through catalog republication. `FallbackConfiguredRepoRefs` matches previously tracked refs by that provenance before the route path, so a provider-side rename plus a transient resolve failure falls back to the tracked ref — keeping its provider id (letting `ExpandedRepoSet` collapse the overlapping-glob duplicate by identity) and its archived flag (no synthetic `Archived: false` ref resuming live polling on reload). Tests: `TestFallbackConfiguredRepoRefsMatchesRenamedRouteByConfiguredPath`, `TestRepoRefFromCatalogKeepsConfiguredRepoPath`, and `TestConfigReload_FallbackKeepsRenamedArchivedTrackedRepo` (mutation-checked: disabling the provenance match fails it). Startup with no prior tracked state still synthesizes an identity-less ref for a renamed exact entry — archived state isn't persisted in the catalog, so there is nothing to recover until the first successful resolve, which self-heals it.

wesm:
6e4909d8 addresses all four findings from the review of e0d44598:

**Provenance lost through dedup.** Glob refs no longer carry `ConfiguredRepoPath` — a pattern identifies no single entry to correlate with — and `ExpandedRepoSet.Add` merges exact-entry provenance across duplicates in both directions (resolved-replaces-fallback keeps the fallback's provenance; a later exact duplicate fills an empty slot). The overlapping-glob rename now retains the exact entry's provenance across repeated reloads. Tests: `TestExpandedRepoSetMergesExactProvenanceAcrossDuplicates` (both orders), and `TestResolveStartupReposPrefersResolvedOverFallbackDuplicates` now asserts the resolved glob ref carries the exact entry's provenance end-to-end.

**Snapshot overwrites newer provenance.** `publishResolvedRepository` now always preserves the currently tracked `ConfiguredRepoPath` under `reposMu` — config resolution is its only author, so provider metadata freshness is irrelevant to it. Test: `TestPublishResolvedRepositoryPreservesNewerConfiguredRepoPath`.

**Tolerant resolver swallowed infrastructure errors.** `resolveRepositoriesTolerant` skips only `*platform.Error` failures (invalid ref, provider not configured, missing capability); store errors and anything else unclassified propagate, so a broken database fails the pass instead of reporting an idle success. Test: `TestRunEligiblePropagatesStoreFailure` (closed store → `RunEligible` errors).

**E2E coverage.** `TestArchiveWorkerSkipsUnresolvableTrackedRepoE2E` drives the worker through the real syncer-as-source wiring with one healthy and one unresolvable tracked repository: seeding degrades per-repo, `/archive/start` via the HTTP API, two `RunEligible` passes complete the healthy repository's issue and merge-request inventories (verified in SQLite and via the status API), and reverting the tolerant resolve makes the test fail.

wesm:
19ab1268 addresses the three findings from the review of 6e4909d8:

**Authoritative publication clobbering a mid-flight archived flip.** `publishResolvedRepository` now preserves the tracked archived value whenever it differs from the operation's snapshot, regardless of `archivedAuthoritative` — a divergence means a concurrent resolution flipped it, and the in-flight provider response cannot be ordered against that flip. Test: `TestPublishResolvedRepositoryPreservesMidflightArchivedFlip`. The existing same-snapshot case still lets fresh provider metadata win in both directions.

**Startup rename duplication.** `fallbackExactFromDB` (mirroring `fallbackGlobFromDB`) resolves a failed exact entry through catalog route history: `ListRepositoryCatalog` matches historical routes, so the fallback recovers the stable provider id and the renamed repository deduplicates against an overlapping resolved glob across restarts. A reused route prefers its current occupant; multiple historical matches with no current occupant fall back to synthesis. Regression test: `TestResolveStartupReposRecoversRenamedExactEntryFromCatalog` (rename observed in catalog, transient exact failure, successful glob → one tracked ref with id, archived flag, and exact provenance).

**Settings mutations erasing provenance.** `mergeTrackedRepos` and `replaceGlobRepos` snapshot provenance from the prior tracked set (`trackedRepoProvenance`, keyed by identity then route) and apply it to incoming refs that lack it — including the glob-refresh path that drops and re-adds refs rather than replacing in place. Tests: `TestMergeTrackedReposPreservesExactEntryProvenance`, `TestReplaceGlobReposPreservesExactEntryProvenance`.

wesm:
41194fdb addresses the review of 19ab1268. One finding fixed, three declined with reasoning:

**Fixed — route-reuse provenance transfer (settings_handlers.go:415).** Route-keyed recovery now refuses to cross conflicting stable provider ids: `trackedRepoProvenance` records the source ref's provider id, and `withTrackedProvenance` transfers by route only when neither side has a conflicting id. The renamed repository keeps its provenance through identity; the route successor inherits nothing. Tests: `TestMergeTrackedReposDoesNotTransferProvenanceAcrossProviderIdentities` (red-first), plus a provenance assertion added to the overlapping-entries glob-refresh handler test (HTTP path, mutation-checked).

**Declined — ordering archived publications by generation/timestamp (sync.go:5591).** This is the generation-counter design declined two reviews ago, re-raised against the narrower residual it left. The remaining window needs two concurrent authoritative resolutions of the same repository racing with a provider-side archived flip between their snapshots; the wrong outcome is a stale boolean that the next resolution (reload, settings refresh, or sync-cycle catalog reconciliation) corrects. Threading an observation generation through every snapshot/publish site is real complexity against a self-healing, vanishingly narrow race — the current heuristic already biases toward the newer write in the plausible single-flip case.

**Declined — durable configured-path→provider-id binding (main.go:851).** The sole-historical-match acceptance is the mechanism the previous review requested for startup rename recovery; a renamed repository's configured path is by definition only a historical route, so accepting only current occupants would undo that fix. The misbinding scenario requires a configured path that was reused by a repository the catalog has never observed, while startup resolution fails — strictly narrower than the rename case it fixes, and it self-corrects on the first successful resolve, exactly like the synthetic fallback it replaces (which syncs nothing usable in the interim). A durable binding table is new persisted state and a migration to shave an already self-healing corner.

**Declined — daemon-restart / settings-HTTP e2e.** Provenance is deliberately not persisted: a restart rebuilds it from config plus the catalog, and `TestResolveStartupReposRecoversRenamedExactEntryFromCatalog` exercises exactly that path against real SQLite — a daemon-restart harness would add process lifecycle, not coverage. The settings-HTTP dimension is now covered by the provenance assertion in the glob-refresh handler test; the downstream e2e consequences (dedup, archived retention through reload, worker tolerance) already have integration/e2e tests from earlier rounds.

wesm:
758c1a92 addresses both findings from the review of 41194fdb:

**Case-sensitive provider id comparison.** The route-reuse guard now compares trimmed ids exactly, consistent with `trackedRepoIdentityKey`. Red-first regression: `TestMergeTrackedReposTreatsCaseDifferingProviderIdsAsDistinct`.

**Composed route-reuse regression.** `TestConfigReload_RouteReuseRefreshThenFailedReloadTracksRenamedRepoOnce` runs the full path: startup resolves the exact entry, the provider renames it and a different repository reuses the old route, a glob refresh lands through the HTTP API, and a config reload with a failing exact lookup follows. The renamed repository stays uniquely tracked carrying the exact entry's provenance; the route successor is tracked without it; no synthetic duplicate appears. Note the direct guard behavior remains pinned by the unit tests — the composed test verifies the user-visible end state, which a reload rebuild can partially launder depending on tracked-set order.

wesm:
2e25067c addresses all three findings from the combined review of 758c1a9:

**High — credential alias missing after catalog fallback.** `resolveStartupRepos` now calls `RegisterConfiguredRepoCredentialAliases` when `fallbackExactFromDB` recovers a renamed route, so the exact entry's repo-scoped credential keeps resolving for the recovered route instead of falling through to owner/host routes. The registration only fires when the catalog recovered a stable provider id (the synthetic fallback keeps the configured route and needs no alias). Red-first test: `TestResolveStartupReposRegistersCredentialAliasForCatalogFallback` asserts `RouteForRepo` on the renamed route resolves to the exact entry's repo route.

**Medium — publication matching by route only.** `publishResolvedRepository` locates its tracked slot via `trackedRepoSlotLocked`: stable provider id first (platform/host-scoped, compared exactly), then the route key — rejected when both sides carry different ids. A stale sync of a renamed-away repository no longer overwrites the route successor, and a publication whose snapshot predates a rename still lands on the renamed entry. Red-first tests: `TestPublishResolvedRepositoryMatchesByStableIdentityFirst`, `TestPublishResolvedRepositoryDoesNotOverwriteRouteSuccessor`.

**Medium — removal ignoring provenance.** `repoMatchesConfig` treats matching exact-entry provenance as a configuration match (host/platform-scoped, before the owner check, since renames can cross owners), and `removeConfigRepos` clears provenance whose entry no longer remains so a stale claim cannot bind a future entry with the same path. Both deletion cases are covered through the settings API: `TestHandleDeleteGlobKeepsRenamedExactEntryRepo` and `TestHandleDeleteExactEntryClearsProvenanceOnGlobKeptRepo` (both red-first).

wesm:
c98c35ab addresses both findings from the review of 2e25067c, plus the CI race failure on that commit:

**Publication keyed on the resolved identity.** `trackedRepoSlotLocked` now keys the stable-id lookup on the resolved ref's provider id — the provider response says whose data the publication carries — so an in-flight lookup through a reused route lands on the tracked successor instead of overwriting the renamed repository the snapshot named. The route fallback deliberately still allows displacing the snapshot's own entry (same tracked id as the snapshot): that is the existing configured-route-reuse semantic, where resolving the route observes a replacement repository and the archive lifecycle pauses the displaced one — `TestSyncRepoReplacementReconcilesArchiveLifecycle` pins it, and a blanket cross-identity discard broke it. What is never allowed is landing on an entry whose id conflicts with the snapshot. Red-first test with both repositories tracked: `TestPublishResolvedRepositoryLandsCrossIdentityLookupOnSuccessor`.

**Provenance matching scoped to provider and host.** `repoMatchesConfigProvenance` now requires the entry's provider and normalized platform host, so removing an exact entry cannot keep provenance alive through a same-path entry on another host or provider. Red-first test through the delete API: `TestHandleDeleteExactEntryIgnoresSamePathOnOtherHost` (github.com entry deleted while a ghe.example.com entry with the same path remains). Inside `repoMatchesConfig` the provider/host guard was already applied by the caller; the path-only check is now a separate helper used there.

**CI `go test -race` failure.** Pre-existing test flaw surfaced by scheduling shifts: `TestHandleUpdateSettingsPublishesPullConfigOnlyAfterPersistence` swaps `srv.cfgPath` to force a persistence failure while the config-watcher goroutine reads it on an unguarded fast path. Both sides now go through `configReloadMu`; verified clean under `-race -count=5`.

wesm:
2c66e0c2 addresses all three findings from the review of c98c35ab:

**Cross-identity archived comparison.** The mid-flight-flip heuristic now applies only when the selected slot holds the repository the snapshot named (`sameIdentity` guard); on a cross-identity landing, authoritative resolved metadata applies and only a non-authoritative publication preserves the successor's tracked state. Red-first test with differing archive flags in both directions: `TestPublishResolvedRepositoryCrossIdentityUsesAuthoritativeArchived`.

**Full-stack route-reuse regression.** `TestSyncReusedRouteResolvingSuccessorKeepsBothReposTracked` drives `syncRepo` through real provider resolution and SQLite catalog reconciliation with both the renamed repository and its route successor tracked: a stale-snapshot sync of the reused route lands on the successor, both tracked entries and both catalog rows survive, no duplication. Mutation-checked — keying the slot lookup back on the snapshot id fails it.

**Provider-scoped provenance coverage.** `TestHandleDeleteExactEntryIgnoresSamePathOnOtherProvider` deletes the GitHub entry while a same-host, same-path entry on another provider remains; provenance clears. Mutation-checked — dropping the provider comparison fails it.

wesm:
9th round: the review of 2c66e0c2's finding is fixed. `sameIdentity` now derives from the snapshot-vs-resolved id comparison, not the selected slot: differing non-empty ids are cross-identity even when the route fallback displaces the snapshot's own entry, so a concurrent archived flip on the displaced repository can no longer stamp the untracked replacement's authoritative metadata. Red-first test: `TestPublishResolvedRepositoryReplacementIgnoresDisplacedArchivedFlip` (displaced occupant archived after the snapshot, replacement resolves unarchived — the replacement's value stands). The non-authoritative branch is unreachable for cross-identity publications — a catalog republication without a fresh resolve keeps the snapshot's identity — so no behavior change there.

wesm:
92a82b4e adds the full-stack regression the review of 27be6f16 asked for: `TestSyncRouteReplacementIgnoresDisplacedArchivedFlipE2E` drives `syncRepo` with a real SQLite catalog and fake provider — the tracked route occupant is archived after the pass snapshots it while the route resolves to an untracked replacement, and the replacement keeps its authoritative unarchived state. Mutation-checked: disabling the cross-identity clause fails it.

wesm:
Addressed the roborev review of 92a82b4 in c9a07834.

**Archived state transitions mishandled during normal sync** — fixed both directions. `runOnce` no longer drops archived refs blindly: `reconcileArchivedRepos` performs metadata-only identity resolution for each archived tracked ref, so an upstream unarchive is observed during the normal pass and the repository rejoins live syncing with fresh metadata (still-archived refs and failed refreshes stay excluded). In the other direction, `syncRepo` stops immediately after identity resolution when the authoritative ref reports archived — before clone fetch, overview, label, and item syncing — relying on the publication having already flipped the tracked flag. Full-stack coverage through `RunOnce` in both directions (`TestRunOnceRestoresLiveSyncForUnarchivedRepo`, `TestRunOnceStopsLiveSyncWhenRepoArchivesMidPass`); `TestRunOnceSkipsArchivedRepos` now has its provider truthfully report the archived repo as still archived, since the pass consults the provider.

**Archived glob expansion does not work for GitLab** — fixed. `PreviewOptions` gained `IncludeArchived`; `ListRepositories` (the configuration-enumeration path) sets it, which omits the `archived=false` query parameter and skips the client-side archived drop in `appendPreviewProjects`. Import previews are unchanged and keep filtering (existing preview tests still pin the `archived=false` parameter). Covered at the client layer (`TestListRepositoriesIncludesArchivedProjects`, asserting no server-side archived filter) and full-stack through `ResolveConfiguredRepoWithRegistry` with the real GitLab client against an httptest API (`TestConfiguredGlobResolutionIncludesArchivedProjects`).

All four new tests were written red-first and each catches targeted mutations (unarchive branch neutered, early-return removed, `IncludeArchived` unset, client-side drop made unconditional). Invariants doc updated: the "GitLab globs cannot see archived projects" limitation is replaced with the enumeration/preview split, and a new invariant covers in-pass archived transition observation. The PR description's limitation note is removed accordingly.

wesm:
Addressed the roborev review of c9a0783 in fb26646.

**Archived metadata refreshes ran before eligibility checks** — fixed. `runOnce` now computes credential-bucket eligibility (`repoEligibility`, the same map that gates dispatch) over the full tracked set before archived reconciliation, and `reconcileArchivedRepos` defers any archived ref whose bucket is ineligible — throttled, reserve-exhausted, or deferred by next-sync-after — without making a provider call. Eligibility is read before the pass refreshes rate-limit snapshots, so a gate that recovered upstream defers the refresh by at most one pass; that trade-off is documented on the function and in the invariants doc. Regression test: `TestRunOnceDefersArchivedRefreshForThrottledBucket` (paused tracker on the archived repo's host; asserts no `GetRepository` call and the archived flag survives), written red-first and verified to catch a weakened guard mutation.

wesm:
Addressed the roborev review of fb26646 in 64b06db.

**Archived-only buckets never gained a cadence gate** — fixed. The pre-reconciliation eligibility map (computed over the selected set while archived refs are still present) is now kept and unioned into the `advanceNextSync` set, with the post-reconciliation dispatch map's values winning for buckets present in both. An attempted archived refresh therefore advances its bucket's next-sync gate like any dispatched sync, and the refresh honors the bucket's throttle factor instead of rerunning every base interval. Consecutive-pass regression: `TestRunOnceAdvancesCadenceForArchivedOnlyBucket` (archived-only bucket with a healthy tracker; two back-to-back `RunOnce` calls make exactly one `GetRepository` call), written red-first (failed with 2 calls) and verified to catch a mutation that drops the archived buckets from the advance set.

wesm:
Addressed the roborev review of 64b06db in 9c955a5.

**Full-stack coverage for the cadence fix** — added `TestScheduledSyncFetchesArchivedMetadataOncePerCadenceE2E` in `internal/server/e2etest`: a real GitHub HTTP fixture (repo + rate-limit endpoints), the production `ghclient` provider client, real SQLite, a host rate tracker, and the server harness. Two consecutive scheduled passes assert the archived repo's metadata endpoint is hit exactly once and the tracked ref stays archived. The passes are driven through `RunOnce` — the scheduled-pass entry point — because the explicit `/api/v1/sync` trigger deliberately bypasses cadence gates (`bypassNextSyncAfter`), so routing the test through the API would never exercise the gate under test. Verified the e2e catches the original bug: with the advance-set union reverted, it fails on the second pass's extra fetch.

wesm:
Addressed the roborev Combined Review of 9c955a5 in ba375be. All four findings were verified real and fixed:

**Published archived flip invisible to the in-flight sync** — `publishResolvedRepository` now returns the ref it actually stored, and `reconcileRepoIdentity` adopts it, so a newer tracked archived flip the publication preserved reaches `syncRepo`'s archived check and stops clone/item syncing. Pinned by `TestReconcileRepoIdentityReturnsMidflightArchivedFlip`.

**Archived refresh outside provider-work coordination** — `reconcileArchivedRepos` registers `PriorityNormalIndex` provider work around each refresh, so an admitted archive request on the same credential is preempted instead of overlapping. Pinned by `TestRunOnceRegistersProviderWorkForArchivedRefresh`, which observes `higherPriorityProviderWorkActive` from inside the provider call.

**Empty snapshot id defeats cross-identity detection** — a resolved id conflicting with the landed slot's id is now cross-identity even when the snapshot carries no id, so a displaced occupant's archived flag cannot stamp its route successor. Pinned by `TestPublishResolvedRepositoryEmptySnapshotIDKeepsSuccessorArchivedState`. Existing route-replacement and displacement tests are unchanged and still pass.

**Archived projects crowding GitLab listing limits** — archived inclusion is now an explicit `RepositoryListOptions.IncludeArchived` set only by configuration expansion; default listings (import previews and the repo-import handler, which filters archived after listing) keep GitLab's server-side `archived=false` filter, which runs before the preview limit. Pinned by `TestListRepositoriesFiltersArchivedByDefault` plus the updated inclusion and full-stack glob tests.

All tests written red-first; each fix verified to fail its test under a targeted mutation. Invariants doc updated for all four. Remaining known limitation (pre-existing, unchanged): GitLab config-glob expansion still flows through the 200-project preview limit, so namespaces above it truncate — that cap predates this branch and now also counts archived projects for config expansion only.

wesm:
Addressed the roborev review of ba375be. Both findings declined with reasoning; no code change.

**Declined — per-repository generation/cancellation for the reload-vs-pass archived window (sync.go:5820).** This is the third variant of the generation/ordering machinery this PR has already declined twice (see the earlier decline comments on archived publication ordering). The residual here is strictly more benign than the two prior variants: when a reload flips archived after the publication, the durable state is already correct — the tracked set says archived and the next pass excludes the repository. The only consequence is that the pass already in flight finishes one content sync against a repository that was live when the pass began. That is not a corruption vector: provider reads of archived repositories remain valid, live sync and archive collection already coexist on live repositories under the same provider-work coordination, and the archive lifecycle pause is keyed on catalog identity, not on whether a live pass completed. Cancelling in-flight repo syncs on reload would thread per-repository tokens through every sync site to shave a self-correcting one-pass window — the same trade declined for the narrower (durable-state) races earlier.

**Declined — full-stack e2e for the preemption and reload interleavings (sync_test.go:10444).** These interleavings are deterministic only through the internal seams the existing tests already use. `TestRunOnceRegistersProviderWorkForArchivedRefresh` observes the provider-work predicate from inside the provider call during a real `RunOnce` against real SQLite; preemption given registered work (cancel-and-wait of a pending archive request) is pinned by the archive-lifecycle suite; the reload/settings archived flip and the resulting skip are covered by the settings archived-flip tests, `TestRunOnceStopsLiveSyncWhenRepoArchivesMidPass`, and the archived-cadence e2e — which already proves through real provider HTTP and SQLite that a tracked-archived ref produces no content fetches. An HTTP-level version of the preemption interleaving would need the same internal hooks to be deterministic; through the public surface alone it is either timing-dependent (flaky) or a restatement of the existing coverage.

wesm:
Addressed the roborev Combined Review of ba375be in c1d575b.

**Archived repos could still receive follow-on detail synchronization** — verified real and fixed at three sites. The detail drain's queue is built before the pass observes an archived transition, so it now skips items at two points: a cheap check on the current tracked ref before any probe/budget/clone spend, and a check after per-item identity verification for refs whose flip is only discovered at resolve time. `syncMRForRepo` and `syncIssueForRepo` (the watched-MR path, notification-linked syncs, and the drain's shared fetch machinery) stop after an archived resolve — except under the archive sync budget (`IsArchiveSyncBudgetContext`), which is exactly the hydration path archived repositories rely on, so archive-budget hydration keeps working.

Regression tests, red-first with mutation verification: `TestDetailDrainSkipsRepoArchivedMidPass` (full pass through `RunOnce` with detail budget, a stale open item seeded pre-archive, and a provider that reports archived mid-pass — asserts zero detail fetches), `TestSyncWatchedMRsSkipsRepoArchivedMidPass` (watched-MR pass stops, nothing persisted, flag flipped), and `TestSyncMRForRepoHydratesArchivedRepoUnderArchiveBudget` (pins the archive-budget exemption — fails if the guard is made unconditional). Invariants doc updated.

wesm:
Addressed the roborev review of c1d575b. Both findings are re-raises of the two declined in the ba375be round; the reasoning there stands unchanged, so no code change.

**Declined — generation/cancellation for the post-check reload window (sync.go:10650, 11011, 11587).** Fourth proposal of this machinery on this PR (declined at cddaf10, 41194fdb, and ba375be rounds). The checks now sit at the correct point — after identity resolution, reading the published state — and what remains is the irreducible window every check-then-act has unless a lock is held across provider I/O or cancellation tokens are threaded through every fetch site. The residual is identical in character to the previously declined variants: durable state is already correct (the reload published the flip; subsequent passes, drains, and watched syncs all skip), and the worst case is one in-flight item or clone completing against a repository that was live when the work was admitted. That is transient, self-correcting, and not a corruption vector.

**Declined — full-stack e2e for detail suppression and hydration (sync_test.go:10552).** Same scope as the e2e declined last round, narrowed to detail endpoints. `TestDetailDrainSkipsRepoArchivedMidPass` already runs the complete scheduled pass (`RunOnce` with the detail budget enabled) against real SQLite with a stale queued item and a mid-pass archive transition, asserting zero detail fetches; `TestSyncWatchedMRsSkipsRepoArchivedMidPass` does the same for the watched path; `TestSyncMRForRepoHydratesArchivedRepoUnderArchiveBudget` pins that hydration still succeeds, against real SQLite. The provider-HTTP delta is transport, not logic, and that transport layer is already exercised for this feature area by the archived-cadence e2e (real HTTP fixture + SQLite, asserting which provider endpoints are and are not hit).

PR #856, opened 2026-08-07
docs: replace transient designs with living architecture

wesm:
## Why

Completed design specs and implementation plans remained beside current documentation after features shipped. Their historical detail could appear authoritative even when the code had moved on, creating a parallel source of truth.

## Outcome

- Current internal contracts live in routed context docs, including consolidated fleet and repository-browser architecture.
- Current maintainer workflow guidance stays in staged Zensical input, while only still-binding rationale remains in ADRs.
- All 98 transient documents are removed, and context validation rejects reintroducing the former documentation tree.


wesm:
No code change is warranted for this finding. The previous context guide did route plans and specifications through `docs/superpowers`, but commits `0ce6dfb3` and `4bdef1a9` intentionally supersede that policy before the burn-down: routed context owns current internal contracts, staged Zensical input owns current workflows, and ADRs retain only still-binding rationale that is not derivable from the implementation.

The explicit acceptance criterion for this change is that no Superpowers documents remain. Restoring the 98 deleted artifacts or removing the structural guard would undo the requested ownership migration rather than correct a defect.

PR #860, opened 2026-08-08
fix: distinguish local sync ceiling failures

wesm:
Provider quotas can remain healthy while Forge blocks background sync at its separate local safety ceiling. The previous generic sync error obscured the actual constraint and recovery window.

- Show the local ceiling spend and reset timing as its own typed sync state.
- Keep provider REST and GraphQL health visually independent from the local ceiling.


wesm:
I am not restoring these four files because this finding conflicts with the documentation policy currently on the default branch. `context/docs-authoring.md` explicitly says not to retain completed plans or design specs after extracting their current contracts.

That extraction is already complete for these artifacts:

- The native-SVG/WebKit constraint is recorded in `context/docs-authoring.md` and points to `docs/screenshots/docs-screenshots.spec.ts::nativeSVGSnapshot`.
- The synthetic Codex terminal-overlay constraint is recorded there and points to `docs/screenshots/docs-screenshots.spec.ts::embedSyntheticCodexTranscript`.
- The implementation and regression coverage remain in `docs/screenshots/docs-screenshots.spec.ts`.

The four deleted files were transient implementation artifacts added by #859 after the living-documentation policy had landed. Restoring them—or adding another dated design record—would duplicate those living contracts and violate the current default-branch policy. No code change is needed for this finding.

PR #862, opened 2026-08-08
feat: hide repositories from the UI via a per-repo gear menu

wesm:
Repositories can now be hidden from the UI without removing them from configuration.

- Each exact repository row in Settings gains a gear menu holding the existing local clone path editor and a "Hide from UI" / "Show in UI" action.
- Hidden repositories drop out of the repository list, repository summaries, the repo typeahead, global repo filters, and the mobile activity selector. Direct links, item feeds, and the Settings page itself still show them, and syncing continues unchanged.
- The preference is stored against the provider-verified repository identity, so it survives renames and is not inherited by a different repository that later reuses the same owner/name route.
- Hiding applies to exact repositories only: glob rows have no gear menu and the API rejects glob targets; repositories not yet verified by their provider return a conflict until a sync completes.

### DIFF
--- a/context/platform-sync-invariants.md
+++ b/context/platform-sync-invariants.md
@@ -68,7 +68,11 @@ combining repository-owned history
   preference on every configuration-change path — startup, TOML hot reload,
   and the DELETE handler (detached from request cancellation): globs can keep
   the repository tracked but expose no visibility controls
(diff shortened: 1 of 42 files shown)


wesm:
Repository visibility action in the repository gear menu:

wesm:
clanking

wesm:
This one spiralled out of control and is going to require a redo

wesm:
Tested and working

PR #864, opened 2026-08-09
fix: allow quota ceiling note to wrap

mariusvniekerk:
The provider quota popover inherits the status bar's single-line text behavior. This caused the local sync ceiling explanation to overflow the fixed-width panel and hid guidance that helps maintainers distinguish the local guard from provider quota.

The explanation now wraps inside its value column while the compact quota values stay on one line. A browser layout regression verifies that the note renders across multiple lines.

<details>
<summary>Validation</summary>

- Full Vitest suite: 4,105 passed, 2 skipped.
- Focused mocked Playwright screenshot capture: 1 passed.
- Frontend formatting, lint, Svelte checks, and Effect diagnostics passed.
- Push hooks passed.

</details>

### DIFF
--- a/frontend/src/lib/components/layout/BudgetPopover.svelte
+++ b/frontend/src/lib/components/layout/BudgetPopover.svelte
@@ -284,6 +284,7 @@
     color: var(--text-muted);
     font-size: 0.9em;
     line-height: 1.2;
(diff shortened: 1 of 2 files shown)


mariusvniekerk:

PR #865, opened 2026-08-09
fix: allow deletion of pre-marker workspaces

mariusvniekerk:
Workspaces created before ownership markers remained valid Git worktrees, but delete and retry operations rejected them after an upgrade. This left workspace records stranded with no recovery path.

This change accepts the live Git registration at the persisted workspace path as cleanup authority. Same-repository replacements follow the workspace lifecycle. Foreign checkouts and symlink targets remain preserved, and retry recreates the ownership marker.

<details>
<summary>Validation</summary>

- Full Go suite: 6,294 passed, 11 skipped.
- Non-mutating lint: 0 issues.
- Push hooks: passed.

</details>


wesm:
ah thanks...I hit this a bunch

PR #871, opened 2026-08-10
Filter Activity by actor

wesm:
Busy Activity feeds make it hard to isolate one person's work. This adds an Actor typeahead that filters by the person shown on each activity event. Candidates stay scoped to the tracked repositories and selected time range, so unrelated feed filters do not make selector choices disappear.

On phones, Activity controls sit behind a compact Filters disclosure. The summary keeps the selected actor visible without adding another permanent control row.


mariusvniekerk:
AFAIK this should not necessarily be needed since author should be just one of the things the filter search bar does already?

wesm:
The filter search box matches more broadly than author, so it currently isn't possible to just see "my PRs"

wesm:
UI/UX here is not good, trying to fix

wesm:
Tested this out, it works but it does not filter based on the top-level Threaded author like I want

PR #872, opened 2026-08-10
feat(activity): reflect live workspace activity across item views

wesm:
Provider events can stay quiet while an agent actively works in a linked workspace. That made the relevant pull request or issue remain buried in Activity, Issues, and Pull Requests even though its tmux session was producing output.

This change treats the latest observed workspace output as an ephemeral recency signal across all three views. It keeps provider history unchanged, preserves stable pagination by sorting item lists in SQLite, and uses one subject snapshot so repository renames, associated pull requests, and direct issue workspace links follow the same identity rules. Threaded Activity can now show a workspace-active subject without inventing a provider event.

The repository-wide short lane encountered host-timing failures in unrelated PTY, server-startup, repository-browser, and command build tests while other repositories were compiling. Each reported failure passed in an isolated rerun; the affected packages and the full internal server package passed in the broad run.

<details>
<summary>Validation</summary>

- Generated API contracts reproduce without drift.
- Affected Go packages and the full internal server package pass.
- Frontend type, Svelte, and Effect diagnostics report zero errors.
- Changed frontend unit tests pass (125 tests); every unrelated full-suite timeout file passes in focused reruns (388 tests total across both batches).
- The affected full-stack Playwright suite passes (32 tests), including real tmux first-observation and event-less Activity behavior.
- Non-mutating Go lint reports zero issues.
- Context structure and public-history private-data scans pass.

</details>

### DIFF
--- a/context/testing.md
+++ b/context/testing.md
@@ -236,6 +236,8 @@ bridge on unmount so assertions do not read stale graph sessions.
 Playwright waits must observe the state consumed by the next assertion. Direct API reads after an optimistic
 mutation wait for its response; rendered assertions wait for rendered state, since route refinement can pair new
 controls with old content (`frontend/tests/e2e-full/utc-maintainer-flows.spec.ts:90`, `frontend/tests/e2e-full/repo-browser.spec.ts:480`).
+Tests that compress browser polling must patch the timer primitive used by the scheduler and await the response
+that carries the expected state (`frontend/tests/e2e-full/ci-dropdown.spec.ts:40`).
 
 Playwright suites with `route.fetch()` proxies must unregister routes with
 `page.unrouteAll({ behavior: "ignoreErrors" })` before page teardown; background refetches can otherwise fail outside the completed test (`frontend/tests/e2e-full/diff-view.spec.ts::mockReviewThreadsOnPreviewMarkdown`).
--- a/context/ui-interaction-contracts.md
+++ b/context/ui-interaction-contracts.md
@@ -12,6 +12,12 @@ semantics.
 - Prevent narrow regressions that usually show up only after review or in e2e
   flows.
 
+## Browser Runtime Constraints
+
+- Do not require secure-context-only browser APIs: remote daemon access may use
+  plain HTTP, so local identity tokens must use runtime-independent identity
+  (`frontend/src/lib/components/terminal/WorkspaceListSidebar.svelte:workspaceRefreshOwner`).
+
 ## Identity And Route State
 
 Interactive surfaces must agree on which item is selected.
--- a/context/workspace-apis.md
+++ b/context/workspace-apis.md
@@ -130,10 +130,46 @@ embedder protocol for arbitrary host state.
 - List/detail reads return persisted plus last-known-good enrichment without
   foreground git or tmux probes; stale components reconcile through bounded
   background workers (`internal/server/workspaceapi/workspace_enrichment.go::toCachedWorkspaceResponse`).
-- `enrichment_status` is aggregate across reads and refresh/push/pull responses:
-  failed reconciliation retains last-known-good components while preserving
-  failure status/error
-  (`internal/server/workspaceapi/workspace_enrichment.go::refreshWorkspaceResponse`).
+- Activity and Issue/Pull Request list and detail responses share one cached
+  subject snapshot: every resolvable subject keeps a workspace reference, while
+  tmux activity remains optional and ephemeral (`internal/server/workspaceapi/subject_activity.go::Handler.WorkspaceSubjectSnapshot`).
+- Workspace enrichment is best-effort on Issue/PR detail: snapshot failures log
+  and omit optional workspace metadata rather than hiding a valid item; list and
+  Activity reads stay fail-fast because the snapshot affects ordering and identity (`internal/server/pullapi/routes.go::Handler.buildPullDetailResponse`, `internal/server/issueapi/routes.go::Handler.BuildDetail`).
+- Threaded and Mobile Activity merge that snapshot after provider-event filters,
+  so workspace recency can keep an event-less subject visible; Flat Activity
+  remains provider-event-only (`frontend/src/lib/views/MobileActivityView.svelte::groups`).
+- Activity normalizes search once and keeps workspace recency for subjects with
+  matching provider events across incremental polls; metadata search filters only
+  eventless rows (`internal/server/huma_routes.go::Server.workspaceActivityResponse`).
+- Activity event references key that snapshot by stable repo ID: issues use
+  own identity and PRs use resolved identity, so route reuse stays fail-closed
+  (`internal/server/helpers.go::workspaceRefForActivityItem`).
+- The shared subject snapshot holds the repository-reconciliation read barrier
+  across both its workspace-summary and subject-metadata reads, so a route move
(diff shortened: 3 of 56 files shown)


wesm:
Addressed all three findings in e665aaa2e.

- Enrichment now tracks divergence and tmux errors independently, aggregates them for responses, and clears only the component that recovered.
- A queued full refresh remains behind an active tmux flight, and unique flight IDs prevent stale completions from releasing newer ownership.
- The real-tmux full-stack test now creates PR- and issue-backed workspaces, proves both APIs rank live work ahead of competing provider-only items, and verifies timestamps, references, and both sidebars.

Verification: `go test -race ./internal/server/workspaceapi`, `go test ./internal/server/... -shuffle=on`, frontend type diagnostics/lint, and the six affected Chromium specs (32 passed).

wesm:
Addressed all three findings in f061dae9f.

- Activity event rows now carry stable `repo_id` and resolve workspace references from `WorkspaceSubjectSnapshot`: Issues use `OwnReferences`, while Pull Requests use resolved `Subjects`. Rename and route-reuse regressions cover preservation and fail-closed replacement behavior.
- “By status” now sorts each Pull Request group with the shared `effectiveActivity` calculation, with a focused ordering regression.
- Aggregate enrichment becomes `fresh` only after both divergence and tmux components complete. Fresh partial results remain `pending`; aged partial results remain `stale`.

Verification: affected Go packages and focused race tests passed; workflow/effective-activity unit tests passed (6 tests); generated API artifacts are unchanged; and the seven affected Chromium specs passed (33 tests).

wesm:
Addressed the mobile Activity finding in `a26a5a812`.

- Mobile groups now merge `workspace_activity`, rank by effective workspace/provider recency, and retain workspace-only PR/issue cards without synthetic timeline events.
- Workspace-only cards keep their workspace indicator and navigate to the phone-focused detail route.
- The full-stack workspace-session spec now exercises the flow with an iPhone 13 profile, including ordering, zero-event rendering, recency labeling, overflow, and navigation.
- The earlier CI failures were also repaired: `PullItem.test.ts` is formatted, and the route-reuse notification fixture no longer expires from the default seven-day Activity window.

Local verification: exact `vp run -w check`; 3,898 unit tests; shuffled `internal/server/e2etest`; the reused-route regression under `-race`; and 45 affected Chromium tests.

wesm:
Addressed the latest roborev findings in `5a37ecb0d`.

- Hold the repository-reconciliation read barrier across both workspace-summary and subject-metadata reads.
- Broadcast first tmux-only completion plus tmux error and recovery transitions, while routine activity movement remains silent.
- Add a real-SQLite API test that verifies Activity, Pulls, and Issues preserve references after a rename and fail closed after route reuse.
- Replace the fixed-delay tmux e2e transition with a two-phase file signal sent only after both baseline observations.
- Clean the testify-helper violations that the newly installed commit hook exposed in earlier feature tests.

Verification includes the full affected Go packages, focused race tests, the complete frontend check, the real-tmux Playwright spec, `make lint-check` with 0 issues, and all installed commit and push hooks.

wesm:
Addressed the latest roborev findings in `5b3972c92`.

- Activity now holds one repository-reconciliation read barrier across the provider-event query and the under-lock workspace-subject snapshot, preventing concurrent renames from mixing route epochs.
- Workspace metadata and Issue/PR effective-activity sorting now use single-parameter JSON-backed SQLite relations instead of three bind variables per retained subject.
- Added a deterministic concurrent-rename API regression plus 11,000-subject DB regressions for metadata, Pull Requests, and Issues.

Local verification includes the affected package suites, the full root server suite, a focused race run, context and testify guards, and two final zero-issue lint runs. The installed pre-push hook also passed golangci-lint and NilAway.

wesm:
Addressed the detail-response finding in `9b8b4670e`.

- Pull details now resolve workspace references from the shared stable-ID `Subjects` snapshot.
- Issue details now resolve their owning workspace from `OwnReferences`, preserving issue navigation when an associated PR exists.
- Removed the obsolete route-keyed workspace-manager dependency from both item handlers.
- Extended the real-SQLite rename and route-reuse test to cover both Pull and Issue detail endpoints as well as their lists and Activity.

Focused package tests, the detail regression under the race detector, the full root server suite, context/testify guards, and the final zero-issue lint gate all pass. The installed commit and pre-push hooks also passed.

wesm:
Addressed the Activity search finding in `c2b9171bc`.

- Search is normalized once, so whitespace-only input behaves as no search on both provider and workspace rows.
- Workspace recency is retained when a provider event matched by author or body, even if the subject metadata itself does not match.
- Incremental polls re-evaluate the bounded in-window provider result set before replacing `workspace_activity`, preventing an existing searched thread from losing its effective ordering.
- Subject metadata search applies only to workspace-only rows.

The focused regressions, focused race run, full root server suite, context/testify guards, final zero-issue lint, commit hooks, and pre-push golangci-lint/NilAway checks all pass.

wesm:
Addressed the missing full-stack search-poll coverage in `4210d355f`.

The new regression uses a real HTTP server, SQLite database, generated API client, and fake tmux process. It seeds a provider event that matches only by author, waits for observed tmux output, performs the initial searched Activity request, then polls with the returned `after` cursor. The incremental response has no new provider items but still retains the workspace subject and reference in `workspace_activity`.

The focused test passes three consecutive runs and under the race detector. The full server suite, zero-issue lint, context/testify guards, commit hooks, and pre-push golangci-lint/NilAway checks also pass.

wesm:
Addressed the component-freshness finding in `ee18bded4`.

A tmux-only result no longer contributes to divergence freshness through the aggregate attempt timestamp. Each component now schedules only from its own attempt and refresh times, so an unattempted divergence probe remains immediately due on the next full workspace read. A focused regression covers the exact tmux-first/full-read-within-TTL state.

Verification: focused red/green regression, adjacent staleness tests, focused race, full workspace API package, full server subtree, zero-issue lint, context/testify guards, commit hooks, and pre-push golangci-lint/NilAway all pass.

wesm:
Addressed the detail-endpoint finding in df1687581.

Pull Request and Issue detail reads now treat the shared workspace snapshot as optional enrichment: failures are logged with the requested item ID, the valid detail response is returned, and `Workspace` is omitted. Activity and Issue/PR list reads remain fail-fast because the snapshot participates in ordering and identity there.

Added DB-backed regression coverage for both detail builders, including a red/green failure-path check. Local verification passed the focused race tests, the full `./internal/server/...` suite, API regeneration, `golangci-lint`, the testify helper guard, context validation, and both commit/push hooks.

wesm:
Fixed the Workspaces route crash in 6fa15ffe.

Root cause: remote daemon access over plain HTTP can expose `crypto` without the secure-context-only `randomUUID` method. `WorkspaceListSidebar` called it unconditionally for a local refresh-owner token and failed before the route rendered. The token now prefers a UUID and falls back to a unique per-page sequence; it does not require cryptographic randomness.

Coverage now removes `randomUUID` in both the component runtime and a full-stack browser session. Local verification passed all 3,899 frontend unit tests, Svelte/type/effect diagnostics, the commit and push hooks, and the complete Workspaces Playwright spec in Chromium and Firefox (26 passed, 6 intentionally skipped live-Kata cases).

PR #873, opened 2026-08-10
Clarify provider quota and app sync limits

cpcloud:
- Show provider quota as capacity left and the app sync limit as spend, with opposite meter direction and accessible descriptions.
- Separate provider throttling from the app's background-refresh boundary and preserve the discovery reserve.
- Treat expired registry and tracker quota windows as unknown at the API boundary so stale capacity cannot drive the UI.
- Show live discovery capacity after background refresh pauses, then identify the hard app-sync ceiling when that reserve is exhausted.
- Collapse credential identity to `host: @user`, keep the single-limit view compact, and label each app limit when multiple credentials exist.
- Fix compact-popover wrapping and meter/value alignment without introducing horizontal scrolling.
- Keep the existing notification route-reuse check inside the rolling activity window instead of pinning it to an expired calendar date.

### Screenshot

Provider quota popover showing remaining provider quota and the app sync limit

### DIFF
--- a/docs/adr/0004-quota-meter-semantics.md
+++ b/docs/adr/0004-quota-meter-semantics.md
@@ -1,0 +1,67 @@
+# ADR 0004: Quota meter semantics
+
+Date: 2026-08-09
+
+## Status
+
+Accepted
+
+## Context
+
+The quota popover presents two different quantities. Provider REST and GitHub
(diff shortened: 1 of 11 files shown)


wesm:
rebasing

mariusvniekerk (approved):

PR #875, opened 2026-08-10
fix: support Windows clipboard in remote HTTP terminals

wesm:
Chrome users who open Forge through a plain-HTTP remote deployment could not reliably paste Windows clipboard text into a tmux workspace. xterm consumed Ctrl-V and Ctrl-Shift-V before Chrome could create a paste event. Copying from tmux also failed because the secure Clipboard API is unavailable on these origins, while Forge's server fallback is intentionally local-client-only.

Forge now leaves native paste shortcuts to the browser and consumes each resulting terminal paste event exactly once. Gesture-authorized terminal copy falls back to the browser's copy event when the modern Clipboard API is unavailable. On insecure origins, right-button mouse events stay out of tmux so Chrome's native context menu remains usable for paste.

The regression serves Forge through a non-loopback HTTP and WebSocket proxy, confirms the page is an insecure context without the async Clipboard API, and uses the real OS clipboard with trusted paste shortcuts. It verifies exact-once tmux input, tmux copy back to the OS clipboard, and no tmux mouse input from right click. The pinned Linux Chromium run exercises Ctrl-V and Ctrl-Shift-V directly.

<details>
<summary>Validation</summary>

- Full frontend Vitest suite: 4,126 passed, 2 skipped.
- Full affected Playwright file: 17 passed, 5 expected browser-specific skips across Chromium and Firefox.
- Pinned Linux Chromium insecure-origin regression: 1 passed.
- Frontend formatting, lint, kit-ui policy, Svelte checks, Effect diagnostics, context structure, privacy scans, and commit/push hooks passed.

</details>

### DIFF
--- a/context/ui-interaction-contracts.md
+++ b/context/ui-interaction-contracts.md
@@ -263,9 +263,13 @@ Keyboard handlers must have one clear owner for each key press.
 - xterm must advertise the Kitty keyboard protocol so terminal applications can
   negotiate detailed key reports instead of misreading legacy cursor input
   (`frontend/src/lib/components/terminal/XtermTerminalPane.svelte::start`).
-- Browser paste shortcuts remain browser-owned inside xterm; Forge consumes each
-  non-empty paste event once at the terminal container
+- Browser paste shortcuts remain browser-owned inside xterm (Ctrl+V on Windows
+  and Linux, Cmd+V on macOS); macOS Ctrl+V remains terminal input. Forge
+  consumes each non-empty paste event once at the terminal container
   (`frontend/src/lib/components/terminal/XtermTerminalPane.svelte::isBrowserPasteShortcut`).
+- Forge-owned paste matches xterm: LF and CRLF become one carriage return
+  before sanitizing and framing; a standalone carriage return stays unchanged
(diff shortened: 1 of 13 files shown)


wesm:
Addressed in `3ddbd5dde`. The regression now runs through a non-loopback HTTP and WebSocket proxy, asserts `window.isSecureContext === false` and `navigator.clipboard` is unavailable, then sends clipboard data through a cancelable browser paste event. It verifies that Forge cancels the event before a later xterm-equivalent capture handler and that exactly one terminal input reaches tmux.

This intentionally starts at the `ClipboardEvent` boundary. Playwright key injection on headless Linux does not make Chromium execute the native browser paste command, even when the clipboard is populated and the xterm textarea is focused. The browser/OS owns shortcut invocation; the Forge regression now covers its full insecure-origin contract from the paste event through the real WebSocket and tmux flow.

wesm:
tested and not working yet, so will keep at this

wesm:
This works now but there is a quirk with how windows \r characters are handled, so taking a look at that to see if there is something reasonable to do

wesm:
This is verified and working

PR #878, opened 2026-08-10
Retain recent terminals for faster workspace switches

wesm:
Sessions with long histories rebuilt xterm, opened a new WebSocket, and replayed buffered output every time a maintainer returned to a workspace. That made workspace switching visibly slower as session history grew.

This keeps recently released terminal sessions in a bounded least-recently-used cache. The default retains 10 sessions, settings allow 0 through 20, and zero restores immediate disposal. Claimed sessions remain presentation state even when their tab is hidden. Released sessions keep parsed terminal state and their socket, but surrender WebGL and leave the cache on disconnect, exit, deletion, generation replacement, or eviction. Retention-limit changes publish only after persistence succeeds, so cancellation and failed saves cannot destroy cached terminals; reversible visual settings continue to preview immediately. This trades bounded memory and background parsing for substantially lower return latency without retaining GPU contexts.

The profiling harness now compares at least 10 retained hits with forced misses at the 64 KiB replay cap. In the local isolated run, ordinary-shell hits painted in 14.0–21.9 ms versus 33.3–63.1 ms for misses; alternate-screen hits painted in 12.7–23.2 ms versus 30.7–83.0 ms. Both scenarios passed the required hit p75 below miss p25 threshold.

The terminal-retention design and plan were temporary workflow artifacts. Their current contracts are distilled into the runtime, UI, config, and testing context, and the repository guard rejects retained Superpowers documents with instructions not to convert them into ADRs. No Superpowers document or feature ADR is included in this pull request.

<details>
<summary>Validation</summary>

- Squashed-commit hooks passed, including the repository short Go suite and frontend checks.
- Focused persistence, settings-component, and session-host coverage passed: 50 tests, including overlapping retention-save reconciliation.
- Focused full-stack Chromium continuity coverage passed: retained A → B → A reuse and zero-retention socket closure.
- Context-sync regression coverage passed: 3 tests, including the no-Superpowers guard and inherited-hook isolation.
- Terminal pool browser coverage passed: 20 tests.
- `go test ./internal/config ./internal/server` passed.
- `make lint-check` passed with 0 issues.
- `make profile-workspace-switch` passed against an isolated backend with 10 hit and miss samples per terminal mode.

</details>

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,6 +92,9 @@ only routes to them.
 ## Git Workflow
 
 - **Commit every turn** — always commit your work at the end of each turn, no exceptions
+- Superpowers design specs and implementation plans are temporary working artifacts; never commit
+  or retain them. Before committing, distill current contracts into the matching `context/` topic
+  docs and delete the artifacts. Do not convert a transient spec into an ADR to preserve it.
 - **Capture context before committing** — before every agent-created Git commit, invoke
   the repository-local `context-sync` skill with `--commit`. Apply clear scoped context
   changes before invoking the normal external commit skill. Block only when an unclear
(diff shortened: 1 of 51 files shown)


mariusvniekerk:
Nice

PR #885, opened 2026-08-11
fix(archive): require complete GitHub merge metrics

wesm:
Archive lifecycle hydration could report a merged GitHub pull request as complete with missing or stale canonical merge evidence. GitHub may expose a test merge commit before merge, while a newer local snapshot can also reject the later canonical merged detail through normal timestamp ordering.

- Require merge time and files-changed count plus fetched canonical head and merge SHAs that exactly match storage before merged GitHub hydration completes.
- Replace a pre-merge test SHA only when repository, pull request, merged-row, and head-SHA identity guards match.
- Preserve merger attribution from canonical detail by using the transaction-current stored merge time when GitHub omits that timestamp.
- Keep hydration retryable when canonical merged evidence is absent or snapshot ordering prevents it from being persisted.
- Reopen rows completed by the prior lifecycle generation so existing incomplete archives repair automatically.

This does not change the database schema or API, and other providers retain their current completeness rules.

### DIFF
--- a/context/github-sync-invariants.md
+++ b/context/github-sync-invariants.md
@@ -450,7 +450,7 @@ fallback repository listing.
 - Archive REST and GraphQL failures must preserve typed authentication and reset-aware rate-limit errors so scheduling defers rather than hot-looping generic retries. (`internal/github/pages.go::archiveTransportError`)
 - GitHub archive code owns historical identity inventory only; hydration must invoke ordinary item sync instead of adding archive-specific lookup, normalization, or persistence. (`internal/github/pages.go::ListIssuesPage`, `internal/github/sync.go::SyncArchiveItem`)
 - Archive item hydration bypasses persisted parent-detail ETags; an unchanged parent representation does not prove that legacy lifecycle timelines are complete. (`internal/github/sync.go::SyncArchiveItem`)
-- A merged GitHub archive lookup requires merge time, merge commit SHA, and files changed; fetched merged evidence survives rejection by newer local detail so hydration remains retryable. (`internal/github/sync.go::requireGitHubArchiveMergedMRMetrics`)
+- A merged GitHub archive lookup requires merge time and files changed plus fetched canonical head and merge SHAs that exactly match storage; absent or rejected canonical evidence keeps hydration retryable. (`internal/github/sync.go::requireGitHubArchiveMergedMRMetrics`)
 - Under repository, number, merged-row, and head-SHA guards, canonical merged detail replaces a stored pre-merge test SHA and fills other missing lifecycle fields without weakening snapshot ordering. (`internal/db/queries_merge_lifecycle.go::FillMissingMergedMRMetrics`)
 - Rejected canonical merged snapshots repair each available lifecycle field and
   the merger event under one route-fence lease; a rejected fenced repair stays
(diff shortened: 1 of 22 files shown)


wesm:
Addressed the combined review in `dbaa4ec4`:

- Archive repair now uses the canonical merged-state predicate, with coverage for a `merged_at`-only GitHub response.
- Post-hydration completeness now queries by the stable repository ID returned from reconciliation, with coverage for a provider-side rename.
- The linked git-dependent CI failure was an incomplete test fixture; it now supplies `files_changed` so the test isolates the intended diff-only failure.

wesm:
Addressed in `1d2b980f`.

- Lifecycle repair now treats merge SHA, changed-file count, and merge timestamp as optional inputs and fills each missing stored field independently. Full-stack archive-report coverage seeds a merged-state-only legacy row with complete diff metrics, omits `changed_files` from GitHub, and verifies hydration completes with repaired `merged_at` and `merged_by`.
- Rejected repair paths now clear the resolved repository ID whenever the route fence changes. `SyncArchiveItem` therefore fails completeness and leaves the item retryable. A deterministic rename-during-fetch regression verifies the actor is not written under stale identity and the archive attempt returns the unresolved-repository error.

Affected full suites, focused race tests, the 6,388-test repository short lane, and non-mutating lint all pass.

wesm:
Addressed in `4e5aa0ef` and `183b1734`.

The rejected-snapshot repair now reads the current stored merge request under the same route-fence lease and uses its persisted `merged_at` to admit merger-event persistence. The database transaction still re-reads the parent and uses the authoritative stored timestamp. Full-stack coverage now combines stored timestamp/metrics with provider `merged_by` and no provider `merged_at`, then asserts the actor in the report.

The linked race-lane failure was a separate test-ordering race. The workspace wire test now observes the initial 0/0 enrichment through HTTP before it mutates Git state and advances the fake clock, so an in-flight initial calculation cannot publish stale data at the new clock time. The full `internal/server/workspacetest` package passed under `-race` after this change.

wesm:
Addressed in `1b0a06d9`.

Accepted merged snapshots now preserve an existing immutable `merged_at` when GitHub omits the field, so a newer partial response cannot erase durable merge timing. The regression test makes the timestamp-omitting response win the update guard and verifies both the stored timestamp and merger event.

Archive verification now also receives an explicit lifecycle-persistence result. A fetched merged pull request cannot complete hydration unless the sync round reaches merger-event persistence. The concurrency test queues a repository rename immediately after the accepted parent snapshot and verifies that hydration remains retryable with no actor event.

The full short suite passed (6,393 tests; 185 expected skips), the focused lifecycle cases passed under the race detector, and lint reports zero issues.

wesm:
Addressed both findings in commit 2b6f12f6.

- Every active merged GitHub lookup completed under an older lifecycle generation is now requeued once, even when its stored fields and actor are populated. Completion at the new generation is the durable proof that canonical merged detail passed the current checks.
- Canonical changed_files now replaces a stale stored value under the existing repository, pull request, merged-row, and head-SHA identity guards. Archive completion also requires the stored count to equal the fetched canonical count.
- The full-stack regression starts with an old-generation row that already has a merge timestamp, actor, stale nonempty test SHA, and stale file count, then verifies canonical repair and report output. Separate coverage verifies a file-count mismatch cannot complete hydration. Responses that omit changed_files still persist an available merger actor but remain retryable.

wesm:
Addressed in commit 9c0f140e.

Archive completion now requires canonical GitHub detail and the transaction-current stored row to agree about whether the pull request merged. A disagreement in either direction returns a retryable `merge_state` mismatch; it does not infer completion from populated metrics or downgrade the newer stored snapshot.

The new full sync-path regression stores a newer merged row with complete metrics, returns an older closed-unmerged canonical response that loses the timestamp guard, and verifies the sync fails closed while preserving the stored state, timestamp, SHA, and file count. The existing inverse case (canonical merged versus newer stored open) now asserts the same state-agreement contract.

PR #887, opened 2026-08-12
fix: support item numbers in Activity search

mariusvniekerk:
Activity search did not match pull request or issue numbers even though the dedicated list views did.

- Make bare (`10`) and hash-prefixed (`#10`) queries match every Activity row for that item.
- Keep notifications and eventless workspace subjects under the same search contract.
- Preserve the existing title, author, body, branch, and commit search behavior.
- Keep boundary-size repository history checks within CI process capacity and give the tmux shutdown probe enough scheduler headroom under the full test suite.


mariusvniekerk:
Activity search filtered to synthetic item #10

PR #892, opened 2026-08-12
fix(activity): reconcile recency after detail sync

wesm:
- Keep Activity recency consistent with PR and issue detail synchronization, including events inserted behind the feed's leading cursor.
- Make full Activity reconciliation latest-successful-request-wins: failed newer reads cannot suppress an older successful reconciliation or strand foreground loading, while newer successful snapshots still replace stale results.
- Add real HTTP/SQLite browser coverage for pull request and issue detail synchronization, including cursor-behind events and reconciliation races.
- Preserve quick-exit PTY status and verify the exact exit code through the HTTP, WebSocket, and SQLite workspace-runtime contract.

### DIFF
--- a/cmd/e2e-server/main.go
+++ b/cmd/e2e-server/main.go
@@ -1779,6 +1779,92 @@ func buildAppState(
 			}
 			return
 		}
+		if r.Method == http.MethodPost &&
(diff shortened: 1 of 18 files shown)


wesm:
clanking

mariusvniekerk (approved):

PR #893, opened 2026-08-13
feat: add reusable repository filter presets

mariusvniekerk:
## Summary

- Restore the shared repository filter on Workspaces and include repositories discovered from reachable fleet hosts.
- Add immutable Global and named custom presets, with create, overwrite, and confirmed delete actions in the selector.
- Persist preset membership by provider-verified repository identity while keeping the active preset or unsaved selection browser-local.

Preset names remain immutable. Overwriting changes only repository membership, and deleting the active preset keeps its repository selection as an unsaved scope.

## Screenshot

Repository selector showing a saved preset, delete control, repository list, and fixed save footer


mariusvniekerk:

PR #894, opened 2026-08-13
Prevent private test tmux servers from leaking

wesm:
Aborted Go test binaries can leave detached tmux servers alive after their temporary socket directories disappear. Those servers keep pseudoterminals open and can eventually exhaust a developer or CI host.

This gives each real private-tmux fixture a recoverable owner identity based on its process ID and start time. Normal exit and termination signals clean the registry. Later test runs can also reap a dead owner when its socket directory is gone. Cleanup accepts only same-user tmux processes under the dedicated test root and never addresses the user's default server.

<details>
<summary>Validation</summary>

- Real tmux recovery tests pass for normal cleanup, termination signals, process-ID reuse, and vanished directories.
- The affected server and workspace packages pass.
- API generation is unchanged, the short pre-commit suite passes, and golangci-lint reports zero issues.

</details>

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,6 +1,7 @@
 version: "2"
 run:
   tests: true
+  allow-parallel-runners: true
 linters:
   default: none
   enable:
--- a/context/testing-basics.md
+++ b/context/testing-basics.md
@@ -23,9 +23,9 @@ fixtures, or changing shell-script coverage.
 - Shell-script tests must execute the script against controlled inputs and
   assert observable output, side effects, or exit status. Do not grep scripts,
   workflows, config, or docs for expected implementation text.
-- Real private tmux tests use `testtmux`: publish gates before runs, remove them only
-  after their runs, and retain startup markers on gate, read, validation, or identity
-  errors; share one stale-owner deadline and never address the default server. (`internal/testutil/testtmux/owner.go::Owner`)
+- Private tmux tests retain markers on gate, read, validation, or identity errors;
+  one deadline covers gate closure and startup draining, and cleanup never addresses
+  the default server. (`internal/testutil/testtmux/owner.go::Owner`)
 - Tmux recovery reaps only definitely absent identities and preserves state when
   lookup is indeterminate; real tmux ownership is supported only on Darwin and
   Linux, which provide high-resolution process starts. (`internal/testutil/testtmux/process_identity.go::processIdentityStatus`)
--- a/context/testing-basics.md
+++ b/context/testing-basics.md
@@ -24,8 +24,8 @@ fixtures, or changing shell-script coverage.
   assert observable output, side effects, or exit status. Do not grep scripts,
   workflows, config, or docs for expected implementation text.
 - Real private tmux tests use `testtmux`: gate server creation under stable
-  PID-plus-start ownership and drain admitted startups before cleanup; never
-  address the default server. (`internal/testutil/testtmux/owner.go::Owner`)
+  PID-plus-start ownership, bound draining, and terminate matching stalls;
+  never address the default server. (`internal/testutil/testtmux/owner.go::Owner`)
 - Use provider live or container fixtures only when fake transports cannot
   catch endpoint or authentication drift. GitHub GraphQL validation is gated by
   `KENN_FORGE_LIVE_GITHUB_TESTS=1`.
--- a/context/testing-basics.md
+++ b/context/testing-basics.md
@@ -23,9 +23,9 @@ fixtures, or changing shell-script coverage.
 - Shell-script tests must execute the script against controlled inputs and
   assert observable output, side effects, or exit status. Do not grep scripts,
   workflows, config, or docs for expected implementation text.
-- Real private tmux tests use `testtmux`: publish admission gates before runs and
-  remove paired gates only after runs; retain PID-plus-start state on gate errors,
-  share one stale-owner drain deadline, and never address the default server. (`internal/testutil/testtmux/owner.go::Owner`)
+- Real private tmux tests use `testtmux`: publish gates before runs, remove them only
+  after their runs, and retain startup markers on gate, read, validation, or identity
+  errors; share one stale-owner deadline and never address the default server. (`internal/testutil/testtmux/owner.go::Owner`)
 - Tmux recovery reaps only definitely absent identities and preserves state when
   lookup is indeterminate; real tmux ownership is supported only on Darwin and
   Linux, which provide high-resolution process starts. (`internal/testutil/testtmux/process_identity.go::processIdentityStatus`)
--- a/context/testing-basics.md
+++ b/context/testing-basics.md
@@ -23,9 +23,9 @@ fixtures, or changing shell-script coverage.
 - Shell-script tests must execute the script against controlled inputs and
   assert observable output, side effects, or exit status. Do not grep scripts,
   workflows, config, or docs for expected implementation text.
-- Real private tmux tests must use `testtmux` with PID-plus-start ownership and
-  an explicit private socket; never address the user's default server.
-  (`internal/testutil/testtmux/owner.go::Owner`)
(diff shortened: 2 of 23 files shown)


wesm:
Addressed all three findings in `2088fea1`:

- Windows `TestMain` paths now skip unsupported real-tmux ownership; all four affected Windows test binaries cross-compile.
- Server test cleanup now uses only the private owner, with a regression test proving cleanup never invokes a default-socket `tmux` command.
- Owner markers are written in a nonmatching staging directory and atomically renamed into the final `run.*` path.

Fresh verification passed for the affected Go suites, focused real-tmux server tests, Windows cross-compilation, the full pre-commit short test suite, and `make lint-check`.

wesm:
Addressed the concurrency finding in `144de0fa`:

- A stable cross-process file lock now serializes validated root setup, stale reaping, and owner publication.
- Owner cleanup removes only its per-run directory; the shared per-user root and lock file remain stable.
- A subprocess regression test proves a second process cannot enter the root critical section early, and a cleanup regression test proves the shared root remains present.

Fresh verification passed for 10 repeated concurrency runs, the full `testtmux` package under the race detector, all affected Windows test-binary cross-compiles, the full pre-commit short Go suite, and `make lint-check`.

wesm:
Addressed the fixture cleanup finding in `c3fa0d3b`:

- Workspace fixture cleanup now receives the fixture configured tmux command and preserves its complete prefix, including the private `-S` socket, for both list and kill operations.
- Cleanup no longer discovers or executes `tmux` from `PATH`.
- A regression test puts a trap `tmux` on `PATH`, exercises list/kill/re-list cleanup through a configured private command, and verifies every invocation retains `-S`.
- Intentionally missing configured commands remain a no-op for PTY-owner fallback fixtures.

Fresh verification passed for the complete `internal/server` package, Windows test-binary cross-compilation, the full pre-commit short Go suite, and `make lint-check`.

wesm:
Addressed the cleanup/startup race in 8cff6f8e.

- Private `new-session` commands now enter an owner admission gate before tmux startup.
- Cleanup closes that gate, drains live startups using PID plus process-start identity, and only then performs the final server kill and process scan.
- Gate state lives under the stable per-user owner root, so cleanup remains safe when the socket run directory disappears.
- A regression test blocks startup while cleanup begins, removes the run directory, and verifies cleanup waits and then kills the server.

Local validation passed: the focused race regression and related cleanup tests 10 times under `-race`, the full affected tmux/workspace/server packages, Windows cross-compilation, repository lint, and context validation.

The linked CI failure was isolated to the unchanged `internal/github/TestResolveDisplayNameDedupsConcurrentLookups` two-second timing assertion; the exact test passed 35/35 locally under the CI race settings. This push has triggered replacement CI so that conclusion can be checked against a fresh run.

wesm:
Addressed the unbounded admission drain in e29ba657.

- Owner cleanup and stale-owner reaping now bound startup draining to the existing cleanup timeout.
- On expiry, cleanup terminates only processes that still match the PID and process-start identity recorded in their admission markers.
- Multiple stalled startups are terminated concurrently, after which the existing final server kill and process scan continue.
- A regression fixture ignores TERM and stalls indefinitely; cleanup escalates to KILL and returns within the bounded window.

Local validation passed: the new regression and related admission/reaping tests three times under `-race`, the full `testtmux` race suite, real-tmux integration tests in localruntime/workspace/server, Windows cross-compilation for all affected packages, and repository lint.

wesm:
Addressed stale-owner cleanup scaling in fd6287c5.

- Stale recovery now validates and closes every stale admission gate before draining.
- All admitted startup identities are deduplicated and drained under one shared cleanup deadline.
- TERM-resistant startups are still terminated concurrently with PID plus process-start validation.
- Admission directories remain fail-closed if the shared drain reports an error.
- The regression creates two stale admission directories with TERM-resistant startup processes and verifies initialization completes within one bounded recovery window.

Local validation passed: the multi-owner regression and related cleanup cases three times under `-race`, the full `testtmux` race suite, real-tmux integration tests, affected Windows cross-compilation, repository lint, and process-leak inventory.

wesm:
Addressed the latest roborev findings in `630740f8`:

- cleanup now preserves both ownership directories when admission closure or drainage fails, so stale recovery can retry safely
- stale reaping records gate status per run, cleans only successfully closed and drained owners, and defers owners first observed alive
- orphan recovery recognizes only the exact `tmux: server (<absolute socket>)` title form, while retaining root-containment and dead-owner identity checks
- regression coverage includes close failure, drain failure, an owner dying between recovery snapshots, and tmux server-title parsing

I also fixed a repository-level validation isolation bug in `2c305063`: golangci-lint now allows parallel runners. The normal pre-push hook passed while an unrelated repository lint was active.

Focused race tests passed three times, the package cross-compiled for Windows, golangci-lint reported zero issues, and context validation passed. CI is running on `2c305063`.

wesm:
Addressed the `2c30506` roborev findings in `6c3edb87`.

- owner publication now creates the admission directory before atomically publishing the marked run
- interrupted publication leaves an admission-only record that stale recovery can close, drain, and remove
- owner cleanup preserves the admission gate when process termination or run-directory removal fails
- stale recovery preserves a paired gate whenever the run remains invalid, unsafe, or otherwise unremoved
- regressions cover publication ordering, admission-only cleanup, owner removal failure, and stale marker-validation failure

The focused race suite passed three times, the full `testtmux` race suite passed, all three real-tmux consumer integrations passed, four affected packages cross-compiled for Windows, and golangci-lint reported zero issues.

wesm:
Addressed the latest process-lifecycle review in `3fa7bc96`.

- Process identity is now tri-state: live, definitely absent, or indeterminate. Stale-owner recovery, startup-marker draining, process inventory, and termination proceed only on definitive evidence; indeterminate lookups preserve ownership state and return an error.
- Darwin and Linux retain high-resolution process-start identities. Other Unix platforms, Android, iOS, and Windows now report real tmux ownership as unsupported instead of using the one-second `ps lstart` fallback.
- Regression coverage verifies that transient lookup failures preserve live owner and startup state and never authorize signaling.

Verification:
- `go test -shuffle=on -race -count=3 ./internal/testutil/testtmux`
- `make lint-check`
- `scripts/context-sync --check`
- commit and push hooks passed

wesm:
Addressed the latest tmux ownership findings in `b264fbd9`.

- Linux `/proc/<pid>/stat` `ENOENT` is now only classified as definite process absence when `kill(pid, 0)` confirms `ESRCH`; otherwise the original lookup error remains indeterminate.
- Invalid or unreadable startup markers are preserved and fail the drain. A read failure is ignored only when `Lstat` confirms that the marker disappeared concurrently.
- Regression tests cover missing procfs with a live process, confirmed process absence, malformed markers, and unreadable markers.

Verification:
- `go test -shuffle=on -race ./internal/testutil/testtmux`
- real private-tmux integration tests in localruntime, workspace, and server
- Linux, Windows, FreeBSD, OpenBSD, and Android compile checks
- `make lint-check`
- `scripts/context-sync --check`
- commit and push hooks passed

wesm:
Addressed the admission-lock timeout finding in `fe21e1fe`.

- Owner cleanup and stale recovery now acquire admission locks with `TryLockContext`.
- One cleanup context covers gate lock acquisition and startup-marker draining, so a wrapper suspended before marker publication cannot block cleanup indefinitely.
- A lock timeout returns `context deadline exceeded` and preserves both run and admission directories for later recovery.
- Subprocess regressions hold the gate lock without publishing a startup marker and cover both owner cleanup and stale recovery.

Verification:
- `go test -shuffle=on -race ./internal/testutil/testtmux`
- real private-tmux integration tests in localruntime, workspace, and server
- Linux, Windows, FreeBSD, OpenBSD, and Android compile checks
- `make lint-check`
- `scripts/context-sync --check`
- commit and push hooks passed

PR #895, opened 2026-08-13
Fix GitHub App merge-method desynchronization

wesm:
## Why

GitHub App repository responses can omit all three merge-method fields when the installation lacks Administration access. The GitHub adapter previously converted those omissions into explicit `false` values, corrupting stored repository settings and causing squash-only repositories to display a plain **Merge** action.

## What changed

- Treat GitHub merge settings as authoritative only when all three fields are present.
- Prefer complete user-authenticated repository metadata while retaining complete App metadata if the overlay is unavailable or incomplete.
- Preserve stored settings on omission so later successful refreshes automatically repair previously damaged all-false rows.
- Hide merge actions when no merge method is known to be enabled.
- Document the complete-field-group retention invariant and cover normalization, overlay, persistence, repair, and UI behavior with tests.

## Validation

- Full pre-commit suite, including short Go tests and frontend checks
- `go test ./internal/github -shuffle=on`
- PullDetail jsdom suite (74 tests)
- `make frontend-check-no-deps`
- `scripts/context-sync --check`
- RoboRev fix audit: no open jobs

A prior non-short `go test ./... -shuffle=on` attempt encountered unrelated host-sensitive timeout/PTY cleanup failures; the subsequent required short suite passed in full before commit.


wesm:
Addressed in e62cf3caa.

- Extended `TestGitHubAppSplitAuthE2E` to seed a damaged all-false SQLite row, verify incomplete App/PAT responses retain it through `/api/v1/repo/...`, then verify a complete PAT overlay repairs the API-visible settings to squash-only.
- Added Chromium Playwright coverage where `ViewerCanMerge` is true but all three merge methods are false; the merge action and modal remain absent.

Validation:
- GitHub App server e2e tests pass.
- All 12 `detail-stale-actions.spec.ts` Chromium tests pass.
- Full pre-commit and pre-push hooks pass.

wesm:
The incomplete-settings/defaults case is an accepted residual risk from the approved design. Known rows retain their last complete settings and later complete refreshes repair damaged rows. Fresh rows can temporarily expose permissive schema defaults if GitHub omits all merge fields while viewer permission is known; an unsupported choice still fails at the provider. We are intentionally not adding merge-settings validity state or changing repository initialization semantics in this fix.

wesm:
Fixed the race CI failure in 3d3aa9080.

`TestKataEffectiveLinkHydrationBatchesAndLimitsPerDaemonConcurrency` used a 25 ms sleep to infer overlap. Race instrumentation could serialize the requests and observe one active request even though the limiter allowed two. The test now uses an explicit barrier: two batches must occupy both per-daemon slots, the third must remain blocked, and then all requests are released.

Validation:
- Focused race test passes with the failing CI shuffle seed.
- Full `internal/server/kata` race package passes.
- Full pre-commit and pre-push hooks pass.

mariusvniekerk:
Review comment seems unhinged.  how would the repository ID change...

wesm:
No code change for the latest RoboRev finding.

The user-authenticated repository read is an existing best-effort overlay for viewer permissions; this PR extends that same overlay to merge settings only when the complete field group is present. Adding a second repository-identity protocol for route reuse between the App and PAT reads is broader than this fix and would need a deliberate policy for retry, omission, and permission handling across the whole overlay—not a merge-settings-only guard.

Persistence remains protected by the repository observation identity/freshness fence. We are accepting the narrow between-read route-reuse risk here rather than expanding this PR into a general overlay-consistency redesign.

wesm:
ignoring, merging

PR #899, opened 2026-08-13
Configure the default workspace sidebar view

wesm:
## Why

Workspaces sourced from a pull request or issue always open the Diff sidebar first. Maintainers who primarily work from the source item have to switch tabs every time.

## What changed

- Add a persisted Forge setting for the initial workspace sidebar: Diff or PR/Issue.
- Open the source PR or issue by default when that preference is selected.
- Keep Diff as the default and fallback for ad-hoc and Kata workspaces.
- Preserve each workspace’s explicitly selected tab over the global initial-view preference.
- Expose the setting in Workspace settings and update the generated API clients.

## Validation

- Full pre-commit suite, including frontend checks, golangci-lint, and `go test ./... -short`
- Focused frontend tests: 142 passed
- Focused config, server, and settings API Go tests
- `git diff --check` and public-history/private-data scan


mariusvniekerk (approved):
Nice

wesm:
Fixed the CI compatibility failures in `df3b53e21`. Workspace view and startup tests now provide the complete workspace settings contract, and the setting uses the shared `SelectDropdown` instead of a native `<select>`. Targeted unit tests pass 50/50 and affected browser tests pass 15/15.

wesm:
Addressed the latest review in 1112e789f:

- Added a generation-aware workspace settings persistence coordinator that rebases field-level mutations onto confirmed authoritative state and ignores stale hydration results.
- Added regression tests for stale hydration and concurrent sibling-field updates.
- Added full-stack Playwright coverage for persisted PR/Issue defaults and per-workspace Diff precedence.

Local verification passed: focused Vitest suite (166 tests), frontend check, Chromium full-stack scenario, context sync, diff check, and all pre-push hooks.

wesm:
Addressed the stale partial-save race in a5230600c:

- Successful workspace saves now confirm and reconcile only the keys owned by that mutation.
- Hydrated or independently updated sibling settings remain authoritative when an older PUT response arrives.
- WorkspaceSettings now republishes the reconciled store state instead of the raw full response.
- Added a regression test for hydration completing before an older save response.

Verification passed: red/green regression, focused store and component suites (6 tests), Svelte analyzer, context and diff checks, commit hooks, and pre-push hooks.

wesm:
Addressed both persistence failures in eba8ebd9f with a focused change:

- If hydration overlaps an in-flight workspace PUT, the coordinator performs one rebased follow-up write so the persisted sibling field matches the authoritative hydrated state.
- Workspace saves receive the pre-edit settings snapshot explicitly, so a SettingsPage-style optimistic update can still roll back after failure.
- The race regression now asserts the persisted server state and compensating request; the failure test mirrors SettingsPage updating the global store.

Verification passed: affected suite (11 tests), Svelte analyzer, full frontend commit check, and pre-push hooks.

wesm:
Pushed `274ab7b80` with the CI fixes and latest `main`:

- serialized settings PUTs with config reloads so the watcher cannot republish a stale activity default
- removed the invalid inline-workspace assertion from the dock-focus e2e; the test now waits on the dock surface it actually exercises
- added a server regression test for the config-reload race

Fresh local verification passed: pre-commit (including frontend/Effect and short Go), pre-push lint/NilAway, the dock-focus test 10/10 in parallel, the activity-settings e2e 10/10 in parallel, and the server regression 20 consecutive runs.

wesm:
Resolved the stale form snapshot finding in 26d71b93f. WorkspaceSettings now renders and builds mutations from the reactive authoritative settings store instead of the page-load snapshot, so hydration remains visible and an unrelated save preserves the newer sibling field. Added a regression test for hydration followed by changing the default sidebar view.

wesm:
Updated in efdb39553: merged current main (including the bounded local hook from #906), preserved both workspace-sidebar and repository-preset settings state, and added scheduling headroom to two load-sensitive polling tests encountered during the merge. The corrected pre-commit suite and pre-push checks pass locally; fresh CI is now running.

PR #902, opened 2026-08-13
Close merge modal before deleting workspace

wesm:
Deleting a workspace after merge can take much longer than the provider merge itself. The merge dialog stayed mounted throughout teardown, so live refreshes could repeatedly remove and restore it. If teardown failed, the transient warning also left no durable recovery state.

This change treats provider merge completion and workspace teardown as separate phases. Once deletion is durably admitted, the merge response closes the dialog while Workspaces shows the workspace as deleting. A failed teardown remains visible as deletion failed and offers retry or confirmed force deletion. A server restart converts interrupted deletion into the same explicit recovery state.

<details>
<summary>Validation</summary>

- Full pre-commit suite passed, including golangci-lint, frontend checks, and short Go tests.
- Focused database, workspace API, pull API, server integration, and frontend regression tests passed.
- Generated API clients and context documentation checks passed.

</details>

### DIFF
--- a/context/deferred-merge.md
+++ b/context/deferred-merge.md
@@ -29,11 +29,11 @@ supersession, completion events, or pending-state presentation.
   merge; queueing a second one returns 409 `already_pending`, so the UI must
   not offer deferred actions while `deferred_merge_pending` is true.
 - Deferred merge requests retain the selected workspace ID; only a successful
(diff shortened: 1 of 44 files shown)


wesm:
Need to do some QA on this one

wesm:
This is tested and working

PR #917, opened 2026-08-14
Isolate Forge sessions on a dedicated tmux server

wesm:
Forge previously shared the user's global tmux server, so heavy workspace and agent activity contended with personal tmux sessions.

- Workspace terminals, agent sessions, and fleet tmux probes default to a dedicated tmux server (`tmux -L kenn-forge`). Every empty-command fallback shares `config.DefaultTmuxCommand`, and an explicitly configured `[tmux] command` line is used verbatim. Sessions started by older versions stay on the global server and are no longer visible to Forge; `docs/configuration.md` covers cleanup.
- The tmux server permanently retains its spawn environment, so every Forge-issued tmux client (creation, attach, probes, kills, fleet inventory) runs with an exact-name, non-secret allowlist environment. Pane processes receive their real environment only through a self-deleting env-file handoff; the base workspace terminal keeps the credential-sanitized full daemon environment plus tmux's per-pane `TERM`, `TMUX`, and `TMUX_PANE`.
- Credential names and the terminal environment are disjoint by construction: config and Kata catalog validation reject `token_env` names that collide with the allowlist. Configured token names, built-in provider/Kata credentials (case-folded on Windows), and Kata catalog `token_env` names feed monotonic strip sets covering tmux, runtime, and ptyowner sessions, updated at boot and on config or catalog reload. Launch environment from API request bodies reaches only panes, never tmux clients.
- Attach commands pass `-E` and normalize the caller environment (`env -u TMUX`, pinned or unset `TMUX_TMPDIR`); fleet SSH wrapping shell-quotes the remote command.
- Wrapper `[tmux] command`s run with the minimal non-secret environment; custom wrapper data belongs in argv.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,3 +49,6 @@ __pycache__/
 /.roborev/
 .pi/superpowers-state.json
 .kata.local.toml
+
+# ptyowner test artifacts
+.sessions
--- a/context/workspace-runtime-lifecycle.md
+++ b/context/workspace-runtime-lifecycle.md
@@ -151,6 +151,45 @@ still exists.
   `internal/workspace/localruntime/manager.go::Manager.Shutdown`).
 - New tmux sessions use the `forge-` prefix; persisted `middleman-` session
   names remain valid and must not be rewritten (`internal/workspace/`).
+- Forge talks to its own tmux server, not the user's global one: every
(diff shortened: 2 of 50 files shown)


wesm:
I need to test this out today

mariusvniekerk (approved):

PR #927, opened 2026-08-17
feat(terminal): paste images into agent sessions

wesm:
Forge terminals accepted only text from the clipboard, so screenshots had to be saved and referenced outside Forge.

- Pasting one or more clipboard images now uploads them to the active local or Fleet workspace and inserts their paths into the terminal in clipboard order.
- Each paste accepts up to four PNG, JPEG, GIF, or WebP images, with a 10 MiB decoded limit per image and server-side content detection.
- Images live under `.kenn-forge/pasted-images`, stay out of Git, and remain for the workspace lifetime. This does not add a separate media cleanup lifecycle.

### DIFF
--- a/context/fleet-architecture.md
+++ b/context/fleet-architecture.md
@@ -33,6 +33,9 @@ peer transports, or remote workspace and session operations.
   non-hop-by-hop response headers instead of translating remote domain failures
   at the hub
   (`internal/server/fleetapi/fleet_proxy.go::Handler.serveRemoteFleetRESTProxy`).
+- Raw adapter handlers do not enforce Huma `MaxBodyBytes`; body-bounded fleet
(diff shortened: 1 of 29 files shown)


wesm:
Addressed all three findings in `072ac9b39`:

- Added a concurrency-safe 100-entry pasted-image quota per workspace and return a conflict at the API boundary.
- Always pin generated paths in the repository-local private exclude, even when tracked rules already ignore them, then verify the effective result.
- Keep uploads concurrent while delivering separate paste batches to the terminal in paste-event order.

The quota boundary, tracked-ignore removal, and inverse cross-event completion cases now have regression coverage.

wesm:
Abandoning since apparently Ctrl-V works already! sorry for the noise

PR #928, opened 2026-08-17
Keep Forge UI available when provider startup fails

wesm:
## Summary

Forge is an offline snapshot of provider data. Before this change, a GitHub token or identity lookup failure during startup terminated the daemon. The daemon then entered a restart loop, so the local snapshot was unavailable whenever GitHub was unavailable.

Provider startup is now best-effort when sync is enabled:

```mermaid
flowchart LR
    A[GitHub unavailable] --> B[Provider bootstrap fails]
    B --> C[Serve cached SQLite snapshot]
    C -. restart after recovery .-> D[Provider sync resumes]
```

The UI stays available with the last successful snapshot. Provider-backed reads and writes remain unavailable until the provider can be initialized again. Explicit no-sync startup errors remain fatal.

## Verification

- Added a regression test for provider identity/API failure during startup.
- The full short Go suite, lint, and repository hooks pass on the branch.
- With GitHub identity requests returning 503, the daemon stayed running and the documented dashboard route returned HTTP 200.

### DIFF
--- a/cmd/kenn-forge/main.go
+++ b/cmd/kenn-forge/main.go
@@ -535,16 +535,23 @@ func run(opts serve.Options) error {
 		providerSources, err = registerProviderTokenSources(cfg, tokenSources)
 	} else {
 		providerSources, err = collectProviderTokenSources(ctx, cfg, tokenSources)
+		if err != nil {
+			slog.Warn(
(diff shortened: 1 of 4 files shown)


wesm:
Superseded by #932, which includes this branch commit and keeps the local archive available only for transient provider failures while preserving scheduled recovery.

PR #932, opened 2026-08-18
fix: keep local archive available during GitHub outages

mariusvniekerk:
Forge is a local-first console, but startup previously required GitHub credential and identity lookups to succeed before it served archived data. A temporary GitHub outage could therefore make the existing local archive unavailable.

This change treats only temporary network, timeout, rate-limit, and GitHub server failures as degraded startup. Forge preserves configured routes and normal scheduled sync, while using conservative host-level personal access token accounting until the next restart. Missing credentials, invalid configuration, and permanent authentication failures still stop startup.

The live daemon was not restarted during verification. The startup decision is covered by focused tests with injected upstream failures.

<details>
<summary>Validation</summary>

- `make test-short` — 6,559 tests passed; 226 skipped.
- `make lint-check` — zero issues.

</details>

### DIFF
--- a/cmd/kenn-forge/main.go
+++ b/cmd/kenn-forge/main.go
@@ -597,15 +597,11 @@ func run(opts serve.Options) error {
 	archiveService.SetMaintenanceInterval(cfg.SyncDuration())
 	archiveService.SetWake(syncer.WakeArchive)
 	syncer.SetArchiveService(archiveService)
-	hostDegraded := degradedHostLookup(startup.degradedProviderHosts)
 	if err := syncer.SetReposWithContextForDegradedHosts(
-		ctx, repos, false, hostDegraded,
+		ctx, repos, false, degradedHostLookup(startup.degradedProviderHosts),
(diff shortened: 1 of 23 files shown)


wesm:
This overlaps with #928, i will let you reconcile

wesm:
looking

wesm:
clanking

wesm:
Closing in favor of a restart from #928's simpler design. In-process recovery on mutable owner/name routes kept generating real review findings (stale identity behind mutation routes, glob re-discovery, archive reconciliation, mint-cache races), and each fix grew the change further.

Superseded by:
- #940 — best-effort provider startup, now isolated per host (#928 plus its one review finding). Degraded hosts serve cached data with no provider clients, so reads and writes fail closed rather than running on unverified state; sync resumes on restart after the provider recovers.
- #941 — the GitHub App token mint single-flight and failure-cooldown work extracted from this branch; it is independently useful and already review-hardened.

PR #1102, opened 2026-09-15
feat: refresh GitHub activity sooner with a shared relay

wesm:
Add an optional shared GitHub activity relay so Forge can refresh changed pull requests, issues, and repository activity between regular syncs.

- Run the relay as a separate small binary from this repository. It shares only refresh metadata; each Forge keeps its own GitHub credentials and saves pending work across restarts.
- Configure the private feed URL in `[relay]` and restart Forge. Feed polling defaults to every 15 seconds and can be changed with `poll_interval`; normal syncing remains a recovery path.
- Click **Relay** in the bottom bar to see recent changes for tracked repositories and connection status. The button is hidden when the relay is off.

Check and workflow webhooks are ignored to keep them from crowding out other activity. CI results continue to update through normal syncing.

The configuration guide and Faster GitHub updates page explain setup and how to check a real update.


wesm:
false positives

PR #1109, opened 2026-09-16
fix: generate frontend API artifacts through Vite

mariusvniekerk:
Generate the API client and schema constraints together during `vp dev` and `vp build`, and regenerate them when the OpenAPI document changes. This restores startup when generated files are missing and removes separate TypeScript generation from Make and Air.


wesm:
test

Issue #1103, opened 2026-09-15
Convert polling test waits to testing/synctest


rodboev:
352 test call sites use Eventually, Never, or time.Sleep to poll for concurrent outcomes. These patterns are flaky under load and slow CI by burning real wall time on fixed budgets.

Go 1.27's testing/synctest replaces polling with deterministic virtual time. Goroutines parked on Go channels, timers, or mutexes complete instantly inside a bubble. The agentsview conversion (PRs 1708-1710, 1727) removed 85% of these patterns and eliminated flakes without widening budgets.

Convert the convertible subset per call site. Leave OS-event waits (tmux, subprocess, network) on explicit budgets and add a bare-budget check to prevent new sub-second polling from landing unchecked.

## Delivery sequence

1. Server and scheduler tests. 35 call sites in `api_test.go`, 37 in `runtime_launch_rollback_test.go`, and the archive scheduler. Highest density.
   PR: #1110

2. Workspace and e2e tests. tmux launcher, pty owner, and workspace tests. Some will stay on budgets due to subprocess waits.
   PR: #1111

3. Federation and sync tests. 12 call sites in `federation_events_test.go`, plus sync and config reload tests.
   PR: #1112

4. Remaining packages and bare-budget check.
   PR: https://github.com/kenn-io/forge/pull/1127
   Convert stragglers, add a checker to reject new unallowlisted sub-second budgets.

Steps 1-3 are open; step 4 is gated on them.

## Adjacent

- kenn-io/agentsview#1702 is the completed precedent.

rodboev:
PR: [#1110](https://github.com/kenn-io/forge/pull/1110)
PR: [#1111](https://github.com/kenn-io/forge/pull/1111)
PR: [#1112](https://github.com/kenn-io/forge/pull/1112)

mariusvniekerk:
The lint-policy branch (PR #1146) disables kennlint `sleeptest` in `.golangci.overlay.yml` so this conversion can land without the new gate failing on the sleeps that are still here.

A scan of `_test.go` still finds 66 `time.Sleep` calls. The largest piles are `internal/github/archive_wake_test.go` (10), `internal/activityrelay/relay_test.go` (7), `internal/server/workspacetest/pty_owner_e2e_test.go` (5), `internal/server/fleetapi/fleet_monitor_gate_test.go` (4), and `internal/server/workspaceapi/agent_handoff_test.go` (4).

[...]

PR #1112, opened 2026-09-16
test(sync): wait on federation and sync signals directly

rodboev:
Federation lifecycle and sync-trigger tests now wait on exact channel signals instead of polling real time.

The lifecycle test uses `testing/synctest`; sync tests keep their existing deadlines and assertions.

Refs #1103


wesm:
this is a test

PR #1115, opened 2026-09-16
feat: deliver relay activity to Forge over a live subscription

mariusvniekerk:
Forge now receives relay activity the moment GitHub reports it instead of checking the relay every 15 seconds. The relay no longer keeps any data: each webhook is reduced to a refresh hint, offered to every connected Forge, and discarded. Anything a Forge misses while disconnected is picked up by ordinary syncing, so the stored log, cursors, and retention window that existed only for resumable polling are removed along with Forge's checkpoint and pending-work tables.

- Forge holds one Server-Sent Events subscription and reconnects on its own using the repository's existing backoff library, one second rising to a 30-second ceiling. The backoff resets only after a stream has stayed up for that long, so a proxy that accepts and immediately drops connections cannot cause a reconnect storm. A stream that goes silent for three keepalive intervals is treated as stalled and reconnected.
- Delivery is best effort. Each subscriber has a buffer of 256 hints on the relay and Forge queues up to 1024 refreshes in memory, coalescing repeats; beyond that, hints are dropped for regular sync to cover. A Forge that stops reading is released after a 30-second write deadline, and relay shutdown ends open subscriptions immediately rather than waiting on them.
- The **Relay** button shows connected or disconnected instead of a last-checked time. The recent-changes list is unchanged.
- Breaking config changes, accepted because the relay is barely deployed: the relay config drops `database` and `retention`, and Forge rejects `relay.poll_interval` with a message rather than ignoring it. Existing configs that set these keys must be edited.
- The private feed proxy must pass streaming responses through unbuffered and keep idle connections open for at least the 20-second keepalive; Tailscale Serve already does.
- A forward migration drops the two Forge relay tables, since the migration that created them is already on main.

A stdlib SSE stream was chosen over an embedded broker such as NATS. One producer, one direction, and tiny at-most-once messages do not justify a second protocol or broker publish permissions, and the fan-out reuses patterns Forge already has for its own event stream.


wesm:
fixing

PR #1130, opened 2026-09-18
perf: improve Forge on slow connections

wesm:
Make Forge more responsive over slow, high-latency connections.

- Reduce startup downloads with on-demand loading and precompressed assets, and keep large API responses compressed.
- Reuse recent PR and issue views and fixed-revision file previews. Preserve unsent comment drafts across reloads.
- Reduce background polling, let slow list requests finish, and keep cached content visible during reconnection.
- Add a persistent airplane mode that pauses automatic sync while keeping relay updates, item visits, and manual Sync available.
- Remove the top-level Reviews page. Load local reviews when their PR workspace panel opens or the user explicitly refreshes it.

Docker-backed review tests and three WebKit cases remain unverified locally because Docker access and required browser libraries were unavailable.

<details>
<summary>UI screenshots</summary>

</details>


wesm:
Screenshots from isolated fixtures.

mariusvniekerk:
Do we have some actual numbers here for what the reduction in size is?

wesm:
I'll work on producing some more evidence.

wesm:
Yes. In a fresh browser session, the list page downloaded **58% fewer asset bytes** and became visible **6.7 seconds sooner** in a simulated slow connection.

| Measurement | Before | With this PR | Change |
| --- | ---: | ---: | ---: |
| Main JavaScript file, before compression | 4.27 MB | 2.53 MB | 41% smaller |
| All assets needed for the initial list page, before compression | 5.78 MB | 3.76 MB | 35% smaller |
| Those assets actually downloaded, with Brotli compression | 1.87 MB | 0.785 MB | **58% less data** |
| Median time until the list appeared | 17.3 s | 10.6 s | **39% less time** |
| Diff-highlighting workers started without opening a diff | 4 | 0 | Loaded only when needed |

I compared clean production builds of `3adec2c` and `7f23dcd`, using the same dependencies and Chromium 151. The timing test loaded the same seeded list three times per build, each in a fresh browser context, with **700 ms latency and 1 Mbit/s bandwidth**. Download sizes were counted directly from HTTP response bodies, including worker downloads. Sizes use decimal MB and exclude HTML, API responses, and network protocol overhead.

The biggest gain is on a cold cache or after an upgrade. Cached reloads downloaded zero asset bytes in both builds. These are simulated conditions, not measurements from a flight or a live Tailscale connection. Splitting the code also increases initial asset requests from 7 to 24; the timings above include that cost.

This reduces the initial download, not the total amount of feature code: the complete uncompressed asset set remains about 19.95 MB. The follow-up in `98cc8f95d` keeps 3.71 MB of precompressed Brotli copies and removes prebuilt Zstd copies, avoiding another **4.21 MB of embedded assets**. That last number is an asset-storage reduction, not a measured release-binary reduction.

